Issue #67, opened 2014-04-02
Feature request: support media query for breakpoint


gseguin:
Instead of a pixel dimension, it would be fantastic to be able to provide a media query as breakpoint. The underlying code would be using [matchMedia](https://developer.mozilla.org/en-US/docs/Web/API/Window.matchMedia) to support that.

simeydotme:
Closing this old issue, I believe this is mostly covered by the `mobileFirst` option now.

klihelp:
Thanks, have to check this
>> I believe this is mostly covered by the mobileFirst option now.

PR #90, opened 2014-04-05
Implemented Media Query breakpoints support

gseguin:
Fixes gh-67


techagesite:
Hey what does all this mean. Are we able to set custom breakpoints? If so how?

simeydotme:
Out of date, closing, `mobileFirst` option was introduced, instead of media-queries.

PR #123, opened 2014-04-10
Make key-navigation work outside the slick-slider

wibron:
This adds an event-listener to the window instead of the list itself.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -747,7 +747,7 @@
             _.$list.on('mouseleave.slick', _.autoPlay);
         }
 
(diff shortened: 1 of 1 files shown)


bivisual:
> If you want this functionality on your implementation, with one slider on the page, bind $('.your-slider').slickNext or slickPrev() to your window key events.

I'm trying to get exactly that functionality. But even my starting point gets overridden by slicks' keyhandler: 

	$( window ).keypress(function( event ) {
		console.log( event.keyCode );
	});

So how do I do it.

PR #139, opened 2014-04-15
Added lazy loading for iframes

Carioca:
For YouTube videos, etc.

### DIFF
--- a/index.html
+++ b/index.html
@@ -244,7 +244,7 @@ $('.center').slick({
 				</div>
 				<pre><code>
 // To use lazy loading, set a data-lazy attribute
-// on your img tags and leave off the src
+// on your img or iframe tags and leave off the src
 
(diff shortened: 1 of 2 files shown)


iamBenCurtis:
Hi, I think my situation is similar. I am attempting to have this pull a variety of posts from a Wordpress install, (posts can be images, youtube videos, quotes, etc) and have them show up on what is essentially a digital sign in front of an office. My plan is to have Slick load, change slide and check if the post format is "video" and if so tell the youtube (which gets a specific playerID from the posts ID) to play and turn off the autoplay function of Slick. Then when the video is done have the video done event tell Slick to turn autoplay back on and then manually advance to the next slide. I just dont know how to properly write that out or which sections to place each part of code. Any help you could provide would be greatly appreciated. Slick, on the whole, has been amazing to work with for this project and I plan on using it again.

PR #173, opened 2014-04-24
add a option for mouse action on dots

jakeauyeung:
hi. sometimes,we need change a action for mouse,like as "click","hover" on dots,so...i push that.:)

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -100,6 +100,7 @@ $(document).ready(function() {
         infinite: true,
         speed: 500,
         fade: true,
+        mouseAction: 'hover',
         slide: 'div',
         cssEase: 'linear'
     });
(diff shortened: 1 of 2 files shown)


baykorhan:
No it isin't poor. I need this option on a news portal. Because we dont want to click by user , just hover and quickly show slides. if user click to dot, we send them to detail page

I'm looking for an answer for this question on all internet. Do you have ?

simeydotme:
i don't think you need a carousel, @baykorhan :)

baykorhan:
Yes you are right about that, I need this option on "Single Item" format. 
Imagine, there is a news portal and including 5-6 different type slider on the homepage and I want to use just one js library.. slick.js is a multipurpose slider already but we just need more a few little options..

Actually, we do this action with a small update in slick.js but active class too late follow mouseover movement and stuck on another dot. So, I'm looking for official solutions or expert opinion. 

Thank you.

PR #316, opened 2014-06-19
record dragging state on Slick object instead of css class

tomas-mikelskas:
Currently dragging state is recorded as a css class on `.slick-list`. The problem with this is that adding/removing this class causes huge performance problems while scrolling (on mobile devices) on slides that have a lot of DOM elements. My hunch is that it triggers reflow on all nested DOM elements.

This pull request remedies the problem by recording state on the Slick object. The only caveat is that there are some styles applied when `.slick-list` has `.dragging` class. So these are lost, but in my case performance is more important.

If you think the lost styles are critical, maybe style definitions could be added via javascript on dom elements that need them. Hopefully that would not trigger reflow.

I'm happy to update the pull request if you think it's worth while.

Kudos for the great library.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -74,6 +74,7 @@
 
             _.initials = {
                 animating: false,
+                dragging: false,
                 autoPlayTimer: null,
                 currentSlide: 0,
                 currentLeft: null,
@@ -1364,7 +1365,7 @@
 
         var _ = this;
 
-        _.$list.removeClass('dragging');
+        _.dragging = false;
 
         if (_.touchObject.curX === undefined) {
             return false;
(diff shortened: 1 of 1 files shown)


donum:
@tomas-mikelskas @kenwheeler Both, `onBeforeChange` and `onAfterChange` are triggered only after the dragging phase when the mouse is released.

Currently, there is no way to tell when the dragging starts. Tomas, I can see where you come from, but this change has huge impact for others and isn't thought through. Some ideas:

- add events onDragStart / onDragEnd
- make the "dragging" state available outside of slick so you can get it's current state
- make the ".dragging" class assignment an boolean option

Also see [pull request 1859](https://github.com/kenwheeler/slick/pull/1859).

What do you think?

PR #335, opened 2014-06-25
Updated responsive/breakpoints option to work with slickSetOption.

tylerpepper:
When working with `.slickSetOption`, I realized that it was ignoring any `responsive` options that were updated and set. This code change manages to fix that along with allowing `.onBeforeSlide` and `.onAfterSlide` to be used in that `responsive` option.

Originally, `.slickSetOption` was completely ignoring the `responsive` options, and `.onBeforeSlide` and `.onAfterSlide` were not being updated on page load, only when the viewport was being resized.

I'm a little iffy on this fix so any feedback would be helpful! Thanks.


simeydotme:
Thanks so much for this, @tylerpepper ... it helped me get moving towards the solution I have at https://github.com/simeydotme/slick/commit/406da2cd9a1715a11f3b043a89de3e19afa3a272 -- I'm going to close this one as it is quite out of date now, wouldn't be able to merge without significate work, and also is missing a couple of key things (as mentioned in my commit notes) :smiley:

Issue #91, opened 2014-04-05
Add bootstrap


kenwheeler:
Add an embeddable script that loads the CSS, min version and assets. Add ability to specify options via HTML data attribute.

To make things easy for our non coder/jquery unfamiliar friends.

PR #407, opened 2014-07-24
Settings using data attribute

AndreSteenveld:
Configuring slick using an attribute on the element, instead of calling it settings as mentioned in issue https://github.com/kenwheeler/slick/issues/91 I called it slick to avoid potential collisions. For the sake of documentation I've also added a little demo to index.html.


TrySound:
+1

kenwheeler:
This is in, but I did it a little differently

PR #409, opened 2014-07-26
Deep clone to copy any events attached to slides

aaronincincy:
Came across this issue when using slick combined with knockout bindings.  The click bindings on the slides worked fine, however the clones didn't retain their events.  This solved the issue.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1286,12 +1286,12 @@
                 for (i = _.slideCount; i > (_.slideCount -
                     infiniteCount); i -= 1) {
                     slideIndex = i - 1;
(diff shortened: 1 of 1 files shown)


4ver:
@kenwheeler Any update on this?

adam-lynch:
@aaronincincy is there any chance you a working example with knockout or any tips? We're having trouble... some bindings aren't working like ones bound to selects.

aaronincincy:
I don't have anything distilled down to an example handy, but if you look at the slider at the bottom of http://www.macyscollege.com/ you can see what I was able to get working.

PR #585, opened 2014-10-03
Ignore package.json for bower installs.

codekipple:
If package.json is present and i attempt to use browserify to require slick from the bower_components folder the package.json seems to confuse browserify and it's no longer able to find jquery to require.

browserify throws this error when package.json is present:-
```
Command failed: Error: Cannot find module 'jquery'
```

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -19,6 +19,7 @@
         "bower_components",
         "test",
         "tests",
+        "package.json",
         "index.html"
     ],
     "dependencies": {


paulwired:
+1

ghettosoak:
Until this gets fixed, I was able to get around this issue by simply replacing the first factory block:
```
(function(factory) {
    'use strict';
    if (typeof define === 'function' && define.amd) {
        define(['jquery'], factory);
    } else if (typeof exports !== 'undefined') {
        module.exports = factory(require('jquery'));
    } else {
        factory(jQuery);
    }

}(function($) {
```
with this: 

```
$(function() {
```

&& of course, make sure that there is **only one** trailing parentheses at the VERY end:

```
});
```
Not ideal, obviously – and I'm sure that this a fairly common issue when going between modular JS loaders – but it gets the ball rolling, so that you can use this killer plugin. :)

AshD93:
+1

0Neji:
+1 - ran in to the same issue but @ghettosoak's fix also worked for me.

oliverbenns:
+1 I also had this issue and @ghettosoak 's fix also worked for me.

Just for some context, jQuery was already being loaded with a cdn on the project, with that part of the page out of my control as I was creating a self contained plugin that could be dropped into a cms page. I didn't want to include jQuery twice.

codekipple:
This issue just got me again :) I hadn't used slickjs and browserify for a while and took me a while to remember why it could not find jquery.

It's safe to merge straight in as it's only modifying the bower.jsons ignore parameters.

cbracco:
+1

im4aLL:
I also had the same problem with using slick with browserify but none of solutions worked for me. Then i take a took into slick.js and changed - 

Before: 
```javascript
(function(factory) {
    'use strict';
    if (typeof define === 'function' && define.amd) {
        define(['jquery'], factory);
    } else if (typeof exports !== 'undefined') {
        module.exports = factory(require('jquery'));
    } else {
        factory(jQuery);
    }

}(function($) {
```

After:
```javascript
;(function ( $, window, document, undefined ) {
```

add the last line - 

Before:
```javascript
}));
```

After:
```javascript
})( jQuery, window, document );
```

codekipple:
Modifying the js is definitely not the way to go. I would be curious to know if people following @ghettosoak answer tried just changing slickjs's bower.json file like in my pull request instead. That's what fixed it for me.

simeydotme:
Can anyone confirm if this is still an issue? -- if so I'll merge. But it sounds like it was a bug with browserify ?

ThiagoDosSantos:
Problem persists

a-p-co:
+1 Problem persists

PR #600, opened 2014-10-06
Fix responsive swiping edge cases

arbylee:
There was an issue with swiping the carousel beyond the number
of slides if not using infinite mode.  If you are using responsive
breakpoints, you could end up in a situation where you have fewer
remaining slides than your slidesToScroll value and attempt to go beyond
the slider.  When this happened, the user was brought back to where they
started.

This commit makes it so that if a user attempts to go beyond the end of
the carousel, they will go to the end of the list instead of back to where
they started.  Similarly, if a user attempts to go past the front of the
carousel, they will be taken to the start of the carousel.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1440,14 +1440,16 @@
 
         _.currentLeft = _.swipeLeft === null ? slideLeft : _.swipeLeft;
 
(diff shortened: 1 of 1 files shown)


leggomuhgreggo:

PR #616, opened 2014-10-08
Prevent initials.touchObject from being modified

arbylee:
_.initials.touchObject was being modified throughout the code because
$.extend was only making a reference to the object.  Anywhere that
_.touchObject was modified, _.initials.touchObject was modified as well.
This change instantiates a new object for _.touchObject each time.

This problem brought difficult-to-debug issues.  For example, when
breakpoints are met and checkResponsive triggers a refresh, the
touchObject after a refresh wasn't empty!  This led to unexpected slides
snapping on mousemove/mouseleave without actually clicking anything.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -108,11 +108,10 @@
                 slideOffset: 0,
                 swipeLeft: null,
                 $list: null,
-                touchObject: {},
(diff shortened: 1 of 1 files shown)


leggomuhgreggo:

PR #626, opened 2014-10-09
add option to turn off lazy load animation

codfish:
i was occasionally running into issues with the opacity animation where the css would get set to 0 after transitioning to it. Then the animation wouldn't occur, or it had happened already, im not sure. Either way I was left with an invisible active image, because the opacity was 0. Didn't have time to figure out why, so I just added this option, which might be nice to have anyway.


leggomuhgreggo:

PR #639, opened 2014-10-13
Fix blurry text in some cases (Chrome)

LostSenSS:
When "transform: translate3d" has not integer value like "-3001.5px" text is looking blurry in Chrome.
This fix it.


leggomuhgreggo:

PR #651, opened 2014-10-15
Vertical mode with different slide heights

noslouch:
I was having issues with slides of different heights in vertical mode.
The track would advance the height of the first slide, but the slides
were different heights, which meant they were out of sync with the
track.

This patch adds a quick check in `setDimensions` for vertical mode, and
if set, finds the tallest slide and sets all slides equal to that
height. It also sets the list height accordingly.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1182,7 +1182,11 @@
                 });
             }
         } else {
-            _.$list.height(_.$slides.first().outerHeight(true) * _.options.slidesToShow);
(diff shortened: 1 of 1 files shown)


simeydotme:
Could you make a JSFiddle for this, to demonstrate it, and also quickly explain why we'd want all slides to be the same height? -- I'd imagine we'd want it to be variable height depending on the height of the next slide?

noslouch:
It's been a little while since I needed this use-case, but IIRC in vertical mode I had an image with a caption area and I wanted to keep the sizes constant so that the position of the caption area wouldn't slide up and down depending on the size of the image.

Issue #541, opened 2014-09-19
Update documentation on slidesToScroll


SeabassNg:
Apparently, if centerMode is set to true, slidesToScroll is set to 1. Either fix it so that centerMode does not affect slidesToScroll or update the documentation, so I don't need to dig through the code to figure it out.

alfechner:
+1 for your change @eschow . Works for my set even with some breakpoints in it. Though in current version of slick the part to adjust looks some different, I removed

````
        if (_.options.centerMode === true || _.options.swipeToSlide === true) {
            _.options.slidesToScroll = 1;
        }
````
on Line 494:496 in `slick/slick.js`. Maybe it helps someone else.

PR #659, opened 2014-10-15
If centerMode is true, no longer sets scrollToShow to 1.

eschow:
Hoping to resolve https://github.com/kenwheeler/slick/issues/541, but would like to confirm that this doesn't affect any other settings.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -457,7 +457,6 @@
         _.$slideTrack.css('opacity', 0);
 
         if (_.options.centerMode === true) {
-            _.options.slidesToScroll = 1;
             if (_.options.slidesToShow % 2 === 0) {
                 _.options.slidesToShow = 3;
             }


befreestudios:
@eschow, @kenwheeler... this fix only works if there are an even number of slides. If there are a number of slides that are not divisible by an odd number it does not behave as expected.

For example try this fix with 6 slides (works), and then try with 7 slides (broken). I'm trying to track down where this is failing, but in the meantime if someone else can figure it out IT WOULD HELP IMMENSELY. THANKS!!!!

eschow:
From my understanding, centerMode only works with an odd number of slides because there's nothing to center in an even set of slides. Correct me if I'm wrong, @kenwheeler. That was just the assumption I made when I was using this library.

befreestudios:
That seems very inconsistent and restrictive. When working with a variable amount of data this would be unacceptable. If centerMode is to work with slidesToMove it should work with both odd and even number of slides. You would have to assume that infinite was true, but if you scrolled (by 3) past the number of slides available, then it should go back to the beginning. ~ Thoughts?

kenwheeler:
Center mode works with any number

befreestudios:
Thats not what I'm seeing, unless I'm missing something.

With @eschow fix and an odd number of slides, it breaks. And without the fix, no matter what number is set for slidesToScroll, it only jumps one. Try applying the proposed fix above, with slidesToScroll && slidesToShow = 3, and different numbers of slides. It seems to break when numberOfSlides is not divisible by slidesToScroll.

kenwheeler:
I'm saying in general, not this fix

befreestudios:
Thanks again for the help and quick responses. True, in general. Although, when centerMode == true && slidesToScroll == 3, this does not work. Is that a valid bug that has been logged? If not, could we consider this a bug?

To that point, I know this fix above is an attempt to fix this, however all I'm pointing out is that this fix does not fix it for -n number of slides, which is what I would expect. I'll still be digging in too and see If I can find a solution, but please keep me informed as well....

befreestudios:
@eschow BTW the title of this issue is totally misleading. There is no property "scrollToShow", it's either slidesToShow, or slidesToScroll. In your fix you remove _.options.slidesToScroll = 1; and then set the slidesToShow so I'm completely sure what you are trying to accomplish.

@kenwheeler To be clear, The issue I'm having is when centerMode == true && slidesToScroll == 3.

eschow:
My apologies. Multitasking is not my strong suit and I filed it while working on something else.

befreestudios:
This will partially fix the issue I'm seeing. If in centerMode, if the slideCount is divisible by the slidesToScroll property, allow it. Otherwise, set slidesToScroll = 1. However, the case where slideCount % slidesToScroll !== 0, is still not accounted for. <- which is what it breaking for me. Pull Request??

kenwheeler:
ok @befreestudios I'm gonna review this and try and get it in the next release. Been busy with the kid and the holidays, just getting back into the swing of things.

befreestudios:
@kenwheeler, awesome. Thank you. I completely understand (have three three kids as well). Thanks again and let me know if you need anything else from me. Thx.

mhulse:
:+1:  on this fix.

When `centerMode` is enabled it appears as though `slidesToScroll` gets fixed to `1`.

simeydotme:
This is a bit of a mess... and very old... I'll close it and try to note down this issue for the upcoming version 2.

PR #660, opened 2014-10-15
include self and parents

danReynolds:
Came across this issue, at the moment if you are a div with the `slick-slide` class directly, then it will look for your parents and not find you. I'd make a fiddle but this is just very obvious and that seems like overkill. JQuery `closest` is IE6+ compatible and resolves this issue in a way comparable to parents: http://api.jquery.com/closest/


leggomuhgreggo:

Issue #676, opened 2014-10-20
Using `asNavFor` and `slideCount <= slidesToShow`


Petah:
When using `asNavFor` and `slideCount <= slidesToShow` `focusOnSelect` no longer works as expected.

Removal if these lines fixes it, but I'm unsure if it will have any other side effects.

https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L1427-L1429

link2pk:
@kenwheeler , Many thanks for the slider.
Can anyone please check [https://jsfiddle.net/link2pk/00ywqbjL/2/](url) why it is not working here?

Thanks

PR #706, opened 2014-10-22
fix using asNavFor when slideCount <= slidesToShow is true #676

caioflores:


leggomuhgreggo:
bump

simeydotme:
The asNavFor has gone through some serious change since this was opened.
If there's still an active issue for it, we'll get around to it soon.

Issue #727, opened 2014-10-29
Breakpoints do not correspond to css media breakpoints.


srussking:
First of all, love the carousel. It is definitely the best one I have tried and believe me I have tried a ton....

The issue I am having is that the responsive method seems to use $(window).width(), but our css guy is pretty annoyed that this doesn't align with his media queries while resizing the screen in the browser. He understands that this is not what will show up on the actual devices but would still love to see the resizing happen the same as his media queries.

Essentially this is the same issue (I think) that maofree was discussing a few days ago, but seemed unable to articulate. https://github.com/kenwheeler/slick/issues/701. 

I was hoping to introduce an option into that responsive method to allow the dev to use window.innerWidth instead of $(window).width() should they so choose, without forking the repo or hacking the code. Something like $('#carousel').slick({ responsiveUseInnerWidth: true, ... }});

I created a test case on jsbin, since the problem is inherently based on the screen size of the browser I found it helpful to see the results in a full screen view which afaik jsfiddle doesn't allow. Hope that is ok.
Code/Edit: http://jsbin.com/zekakuberu/1/edit?html,css,js,output
Fullscreen Output: http://jsbin.com/zekakuberu/1/

I am planning on developing a pull request which outlines this fix, but figured this was the best place to start.

-Russ

PR #728, opened 2014-10-29
Allow for dev to choose innerWidth

srussking:
For https://github.com/kenwheeler/slick/issues/727

Code/edit: http://jsbin.com/yumiku/2/edit?html,css,js
Full screen: http://jsbin.com/yumiku/2/

This might be better to be able to pass in a function to use rather than a true/false, but perhaps this is enough for folks. Sorry about the excess commits was editing/testing in the browser... ugh.


simeydotme:
Closing for now, We'll try to consider this issue (screen widths / scrollbar calculations) in 2.0.

PR #748, opened 2014-10-31
Adding lazyLoadAttr to make lazy attribute dynamic

noppanit:
I would like to make <code>data-lazy</code> dynamic. So, this option can be set from the options. I have a use case where <code>data-lazy</code> has been used for other stuff already. Is this ok?


simeydotme:
@noppanit this is a decent contribution.
If you could, would you mind rebasing and fixing this up a little so the object's two new properties are both alphabetical, and also check my comments about the variable: `key`

noppanit:
I'll do that and push. :-)

noppanit:
@simeydotme Can you put the comments about the variable `key` again? It might have been deleted.

simeydotme:
@noppanit ... I'm sorry, but I don't even need to test that to see it wont work... however, here's a fiddle to prove it: http://jsfiddle.net/3Ls45p4t/3/ ... the issue is primarily with your placement of `lazyLoadKey` ... the fact that this was never tested, and has been pushed two times both with similar bugs in it... I've lost the confidence to merge. :disappointed: 

I do think it's a decent proposal, though, and will leave it tagged for consideration in 2.0 :)

Apologies. And don't let this shake your confidence... just remember this (and other open source projects) are APIs used by thousands, if not millions of people so need to be tested thoroughly... even a small bug can leave a lot of people high-and-dry :cry: 

Cheers, and thanks for contributing. Simon.

noppanit:
Oh it's my bad that i didn't test it. I will make another PR for this. :)

matiwinnetou:
I have seen already three people that tried to contribute and add this attribute, guess what, I would like to open one as well... :)

Could you please fix this up and rebase so this can be merged in and released :)

If not I will open a new PR as 4th person that needs this :)

We have like 3-4 slick instances in our DOM, when then should or should not get loaded this attribute will allow us to control

matiwinnetou:
+1 on this PR when it is fixed.

PR #781, opened 2014-11-12
775 : for centerMode and slideToShow 1, updateArrows does not apply slic...

robinmalgoire:
...k-disabled


simeydotme:
Out of date...

PR #815, opened 2014-11-19
Make each setting and method linkable in the documentation

jory:
Just to make it a little easier to point out specific things in the documentation.


kenwheeler:
I'm gonna close this because its preventing me from killing the 1.3.12 branch. I actually want to add it though, so I'll work that out.

PR #823, opened 2014-11-24
Update slick.js

dldean:
Fix bug with previous and next buttons when slickGoTo is used to go to a specific slide, add getCount method to return total slide count

http://jsfiddle.net/dldean/v3b4rzhm/


simeydotme:
No longer an issue after other PRs and BugFixes :)

PR #844, opened 2014-12-03
Alternative loadImage method for the lazyLoad functionality

rthijssen:
A while back I wrote a different method for loading images. Rather than using the `` tag I use a `<div>` tag. The image (still specified in data-lazy) is loaded into the divs background. This allows me to center and zoom the background image to fit the size of the div.

  * It is important that the dimensions of the div where specified otherwise nothing will show.
  * The div should also have the class `autoFitImage`

This allows Slick to consistently display a carousel containing images with different dimensions without having to stretch them. I wrote this because I'm dynamically generating carousels from images provided by the user(s).

Example of a slide:
```
<div>
	<div class="autoFitImage <class-with-div-size>" data-lazy="<img-src>"></div>
</div>
```

Small note; I never really worked with scss so not sure if I added the autoFitImage class correctly. I didn’t compile the scss file since it came back containing more changes than just mine. Think you might be using a different scss compiler. The file needs to be compiled before my .autoFitImage class is added to the .css file.

Not sure if you are interested in this method, or if it is appropriate for the project. But thought I should share it.

If you're not happy with this method an alternative solution would be to make the `loadImages` function an option so it is over writable. Let me know what you think.

### DIFF
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -1,4 +1,3 @@
-@charset "UTF-8";
 /* Slider */
 .slick-slider {
   position: relative;
@@ -24,7 +23,7 @@
   .slick-list:focus {
     outline: none; }
   .slick-loading .slick-list {
-    background: #fff url("./ajax-loader.gif") center center no-repeat; }
+    background: white image-url("ajax-loader.gif", false, false) center center no-repeat; }
   .slick-list.dragging {
     cursor: pointer;
     cursor: hand; }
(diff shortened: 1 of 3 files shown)


MentalGear:
Overwriting the load Images functions sounds pretty descent !

simeydotme:
Apologies for the delay on this, but it's a little out of date and not sure the correct approach.
Please be encouraged to open a feature-request issue for lazy-loading background images for version 2.0.

Best, Simon.

PR #847, opened 2014-12-03
add option to allow prevention of autoplay from reversing slide direction

codfish:
A number of sites will typically advertise related or "up next" news items/galleries/etc. after the final slide, and that was the use case I was considering when writing this update. The only way I could think of to do this is currently is put a check in the onAfterChange callback to see if the user tried to advance past the last slide, and if so, display an interstitial of sorts. However, when in autoplay mode, this would never happen because the direction would reverse. I just wanted to add this option so that the carousel would attempt to advance past the last slide, and I could check for that condition and handle it however I wanted, also being able to pause it in that state as well.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -42,6 +42,7 @@ Option | Type | Default | Description
 accessibility | boolean | true | Enables tabbing and arrow key navigation
 autoplay | boolean | false | Enables auto play of slides
 autoplaySpeed | int  | 3000 | Auto play change interval
+autoplayChangeDirection | boolean  | true | Change direction when reaching the final slide when autoplay is enabled and infinite looping is disabled.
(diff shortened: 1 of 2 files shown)


balajigumpena:
From the files changes tab i could find that a new option is added:
+autoplayChangeDirection | boolean  | true | Change direction when reaching the final slide when autoplay is enabled and infinite looping is disabled.


The above changes are not available in the source code. Was this intentional?

codonnell822:
> The above changes are not available in the source code. Was this intentional?

Hey @balajigumpena, this code isn't merged into the source code yet, it's still an open PR.

simeydotme:
Reading the code, I can't see how this actually could work... it looks like as soon as you get to the last slide, it'll just keep going in to oblivion?

Can you create a JSFiddle with a use-case, there's a great boilerplate JSFiddle in the contributing.md file :)

Thanks, Simon.

PR #862, opened 2014-12-12
Add a swipe angle to tune the page scrolling interaction on touch devices.

pixelthing:
Firstly, thanks for an awesomely documented, feature rich, cross-browser, bullet-proof script.

This is a pull request that attempts to improve a tiny aspect of slick's UX.

On a touch device, when you scroll the page up and down, slick carousels often move a little left or right, as you rarely move your finger/thumb/stubby digit in a purely vertical motion. This gives rise to a slightly disjointed scrolling experience. It's because slick sees any movement 45deg from the horizontal axis as being targeted at the carousel (discarding it as a possible page scrolling action), and it's easy to stray into this zone as you move your thumb in mostly a circular motion.

This tweak allows the angle that slick targets to the carousel to be tweaked. In this commit, it's set at the same 45deg as in the current implementation to prevent any difference from current expectations, but I find that an angle of around 20deg gives a better user experience for scrolling the page, without detracting from swiping the carousel left and right. The "right" angle might vary by implementation or target device - hence making it configurable.

I hope you find this useful (even if not complete).

(apologies for the version number bump - it's only so _this fork_ could be given a new github release so bower in our project could update it properly)

### DIFF
--- a/component.json
+++ b/component.json
@@ -2,7 +2,7 @@
     "name": "slick",
     "repo": "kenwheeler/slick",
     "description": "the last carousel you'll ever need",
-    "version": "1.3.15",
+    "version": "1.3.16",
     "keywords": ["ui", "jquery", "carousel", "responsive", "slider"],
     "dependencies": {
         "component/jquery": "*"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "slick-carousel",
-    "version": "1.3.15",
+    "version": "1.3.16",
     "description": "the last carousel you'll ever need",
     "main": "slick/slick.js",
     "repository": {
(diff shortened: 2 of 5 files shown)


simeydotme:
@pixelthing ... this looks like a decent contribution.
Do you think you're able to update it by rebasing, adding in some documentation to `readme.md`... and providing a JSFiddle based on the [contributing guidelines](https://github.com/kenwheeler/slick/blob/master/CONTRIBUTING.markdown) ?

Cpt-Catnip:
Any movement on this PR? Something we'd love to see implemented.

MarcJandt:
@pixelthing Currently I'm using your code the other way around, increasing the angle. I think in this case we need to adjust the following snippet as well to prevent overriding the swiping with a scroll action:
`        if (!_.options.verticalSwiping && !_.swiping && verticalSwipeLength > 4) {
            _.scrolling = true;
            return false;
        }`
Maybe we can use the angle value here as well.

Issue #865, opened 2014-12-13
Add ability to use asNavFor for many slicks


omnilight:
Currently only one slick supported as navigation for, but if you change method to this:

```js
Slick.prototype.asNavFor = function(index) {
        var _ = this, asNavFor = _.options.asNavFor != null ? $(_.options.asNavFor) : null;
        asNavFor.each(function() {
            $(this).getSlick().slideHandler(index, true);
        });
    };
```

it would be possible to use multiple selectors in asNavFor

jfalotico:
You need to add an if check to make sure its not null! 

 if(asNavFor != null) {
         .....
 }

omnilight:
That's right

lsatdown:
I see this works, however it's not been added to 1.4 build?

mothershipsoft:
Is this working in 1.5? Any work around for multiple slicks?

chrisharrisonkiwi:
How do we send the multiple css selectors to Slick?
I attempted:
```
asNavFor:   ['.hole-carousel','.hole-number-carousel'],
```
But then received the error:
```
Uncaught TypeError: $(...).getSlick is not a function
```
So close, yet so far...

chrisharrisonkiwi:
Patched line 344 of: slick.js:

```
var _ = this, asNavFor = _.options.asNavFor != null ? $(_.options.asNavFor):null;
if(asNavFor != null){
	asNavFor.each(function(item){
		var linked_slick = $(asNavFor[item]).slick('getSlick');
		linked_slick.slideHandler(index, true);
	});
}
```

anneminford:
I've got 2 instances of a slider on a page, they work perfectly apart from the dots only appear on the second slider? Any ideas?

jQuery('.slider-for').each(function (idx, item) {
    var carouselId = "carousel" + idx;
    this.id = carouselId;
    var newClassFor = '#'+carouselId +'.slider-nav';
    jQuery(this).slick({
      slidesToShow: 1,
      slidesToScroll: 1,
      arrows: false,
      fade: true,
      asNavFor: newClassFor
    });
});

jQuery('.slider-nav').each(function (idx, item) {
    var carouselId = "carousel" + idx;
    this.id = carouselId;
    var newClass = '#'+carouselId +'.slider-for';
    console.log(newClass);
    jQuery(this).slick({
      slidesToShow: 3,
      slidesToScroll: 1,
      asNavFor: newClass,
      dots: true,
      arrows:true,
[...]

PR #866, opened 2014-12-13
Update slick.js to support multiple selectors #865

omnilight:


simeydotme:
Already added in another code commit.

PR #874, opened 2014-12-17
Less port

andyinabox:
I ported the .scss file into a .less version. It doesn't maintain all the functionality (mainly the stuff contained in functions or conditionals) but it does use the same vars to make it easily customizable in a less project.


kenwheeler:
Closing, as I've redone all the CSS

kenwheeler:
Thanks though

Issue #883, opened 2014-12-22
shouldn't we have to avoid half-pixel translations?


theeluwin:
according to some articles here: http://martinkool.com/post/27618832225/beware-of-half-pixels-in-css, http://www.reddit.com/r/programming/comments/wv731/beware_of_half_pixels_in_css/, I'm suggesting that we should round up or down on codes that uses `translate` or `translate3d`, for example: https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L293
what do you think? can be seen in simple fiddle: http://jsfiddle.net/t0o6zy2j/

> with half-pixels
![2014-12-22 8 52 39](https://cloud.githubusercontent.com/assets/1303549/5524703/01b80c3e-8a1d-11e4-94d3-a722cbbe98c6.png)

> without half-pixels
![2014-12-22 8 52 55](https://cloud.githubusercontent.com/assets/1303549/5524704/040fd94e-8a1d-11e4-95bc-948dff84b5c8.png)

PR #886, opened 2014-12-22
Avoiding half-pixel translations.

theeluwin:
Simple implementation of #883.
Note that the changed codes were meant to affect only the values that has to do with `translate` or `translate3d` in css, since the issue was about half-pixels in translations.
Related article: http://martinkool.com/post/27618832225/beware-of-half-pixels-in-css

p.s. Is it okay to submit only `slick.js` file? How should I treat with the `slick.min.js`?


Spone:
This should be merged :+1:

Issue #361, opened 2014-07-09
Need onChange setting ( not onBeforeChange or onAfterChange )


msavov:
I have n slides and 5 visible slides. I want to change something on the first active slide at one moment. I try with these two methods but this is not what I need. 

I achieve this on clicking the arrows for example: 
$('.slider button').click(function() { 
      $('.slick-active:first').css({
                'background-image': 'none',
        });
});

But when I use autoplay, there isn't click event so I need to call this everytime when active class is set to another set of elements, and then take the first child.

PR #891, opened 2014-12-30
added OnChange support

Geoffrey-D:
I just added the OnChange support as discussed here https://github.com/kenwheeler/slick/issues/361#issuecomment-50963142 a few months ago =)


kenwheeler:
This is in 1.4 as an event

Issue #834, opened 2014-11-27
Option to select how many images to preload with lazyload


alancwoo:
The only feature I am wishing for would be the ability to choose how many images to preload like the `numImagesToPreload` feature in Royal Slider (referring to [royal slider documentation - options - numImagesToPreload](http://dimsemenov.com/plugins/royal-slider/documentation/#options)).

A page I'm working on will have many carousels, around 10-15, with 5-10 images each. On Demand lazy loading helps, but means a delay every time the user changes the image.

Progressive lazy loading means a decent first load of the initially visible images, but a lot of bandwidth used otherwise for images that may never actually be seen.

(Otherwise I am of course so happy with slick, it's an amazing library, many thanks)

kdesigns:
I needed this function too but because SlickSlider doesn't support different sources for 72dpi and retina images I had to use another plugin. I chose lazyLoadXT (https://github.com/ressio/lazy-load-xt).
I implemented it using:
```
slickSlider.on('afterChange', function() {
    var activeSlide = $('.slick-slide.slick-active');
    activeSlide.next().find('img[data-srcset]').lazyLoadXT({
        show: true
    });
});
```

Djules:
There is another value for lazyLoad option in the source code, not documented though: '**anticipated**'.

It seems to do what the OP was looking for.

jamesinealing:
From what I have seen 'ondemand' on an autoplay slideshow always seems to pull in required images about 2 slides ahead of when they will be used, which feels like it handles this in a pretty sensible way

Luc45:
Any news about this? (sorry for duplicate)

pminf:
@kdesigns In version 1.9.0 there is no function called `lazyLoadXT`. Is there a new way to lazy load the next image?

janumedia:
@pminf have you tried `slickNext` ?

`$('.your-element').slick('slickNext');`

bYemma:
> There is another value for lazyLoad option in the source code, not documented though: '**anticipated**'.

Why is this not documented? Works like exspected and was introduced in [release 1.7.1](https://github.com/kenwheeler/slick/releases/tag/1.7.1)

ThanhDv-Rk:
if you wanna preload 1 img, you can use this way 
  let currentS = $(".photo_viewer_slideshow");
  currentS.slick({
    lazyLoad: "ondemand",
    slidesToShow: 1.000000001,
    slidesToScroll: 1,
  });

PR #893, opened 2014-12-30
Add the nextImageToPreloadOption

juli3773:
Hi!

In response to the feature request #834 I tried to purpose a nextSlidesToPreload option.
I had a specific issue on my project that I illustrated here in this JsFiddle with the new option: http://jsfiddle.net/dmhzc360/1/

I don't know if this is the best way to do that but it respond to my need and anyway, i would be glad to discuss other options!
Thanks a lot for this plugin!
Juliette


vyacheslav-lonschakov:
It's really good changes, at now i have white space while slider moves. Why this changes is not merged?

simeydotme:
This is a bit too out of date, and could cause a bunch of regression.
We'll look at trying to enchance this feature when tests are available, and/or in 2.0.

Apologies for the delay and thanks for the effort.

PR #916, opened 2015-01-11
Add lazyLoadSourceAttr and lazyLoadTargetAttr

hultberg:
Allow to change where lazyLoad gets image src (data attribute) with lazyLoadSourceAttr also allow where lazyLoad will put the image (default src on img-tag).

I have added an extra sourceAttr support for one of my projects where we use divs for images to have them floating inside fixed height. If lazyLoadTargetAttr is 'background-image' then lazyLoad will look for divs with matching sourceAttr and simply set background-image on them.


hultberg:
http://jsfiddle.net/47bn325b/

CHEWX:
Good feature, I guess it's not going to be merged in ?

@hultberg

bennybee:
Would be a very useful feature, and would be good if it didn't rely on the container being a div (eg be able to add the data-lazy attribute to an LI tag)

huan086:
Instead of `$('div[' + _.options.lazyLoadSourceAttr + ']', imagesScope).each(function () {`

I used this

                var lazyAttribute = '[' + _.options.lazyLoadSourceAttr + ']';
                $(lazyAttribute, imagesScope).addBack(lazyAttribute).each(function () {

so that I could use A instead of DIV with lazy loaded background. The `addBack` is due to A itself being the `slick-slide`, instead of the usual case of being the children of `slick-slide`.

hultberg:
Thats a good point @huan086 !

stha:
Is this going to be merged within the next time?

hultberg:
@stha: I would need to fix the conflicts first, but I could update this PR and hope for it to be merged.

ahmadalfy:
Please fix the conflicts if there is any ...

Pinging @kenwheeler

masi:
#1661 is related. If I understand #1661 correctly the change would affect all images.

Anyway, what I would require is that I can choose on a per image basis if the src or the srcset ist to be lazy loaded.

yivo:
Are there any progress on it?

gino8080:
dead? :)

vitekvil:
when rows is more 2 after resize of window loaded images are lost and nothing is shown

leofrozenyogurt:
revive?

Issue #637, opened 2014-10-12
Lazy load isn't preloading images beforeahead


dmitry:
Lazy load on demand isn't preloading next image. You can try it out few times using the incognito mode in chrome or firefox.

http://jsfiddle.net/qjqhmtb5/9/

PR #934, opened 2015-01-18
Explain lazyload parameters

peschee:
Added explanation for lazyLoad parameters, since the difference between `ondemand` and `progressive` were not explained in the docs. Found an explanation in #637


tomhicz:
For convenience, I would recommend adding this info to the explanation of slick settings on the main slick web page, because it took me ages to find this.

PR #950, opened 2015-01-24
Adding in page-visibility detection for pausing/resuming carousel (if autoplay)

### LINKED ISSUE #927: Variable height slider bug
Hi,

I found a bug!

I have a horizontal slider with variable height which has autoplay. If I let this play on in a tab which isn't active and then go back, the queue is filled up with the height-animations and they all play, which generates a weird pulse.

You can easily fix this by adding http://api.jquery.com/stop/ to the animation I guess? But don't have any idea on how to do this myself.

Any help is welcome!

simeydotme:
should fix #927 - also adds nice feature in general. 
http://jsfiddle.net/hwj2nnm6/1/

Sorry I don't have all the different possible modes/features in that fiddle, @kenwheeler - I think the impact of the code is pretty contained and minimal - I _did_ read the CONTRIBUTING.markdown :stuck_out_tongue_winking_eye:

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -124,6 +124,7 @@
             _.breakpoints = [];
             _.breakpointSettings = [];
             _.cssTransitions = false;
+            _.hidden = "hidden";
             _.paused = false;
             _.positionProp = null;
             _.respondTo = null;
@@ -132,6 +133,7 @@
             _.$slidesCache = null;
             _.transformType = null;
             _.transitionType = null;
+            _.visibilityChange = "visibilitychange";
             _.windowWidth = 0;
             _.windowTimer = null;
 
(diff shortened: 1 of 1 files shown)


simeydotme:
oh, by the way, I could ___not___ get Make to run on my machine, piece of $hit!! 
My next contribution (if you dont get there first) will be converting this sexy monster of a plugin to Gulp! :stuck_out_tongue: - so long as that's acceptable?

kenwheeler:
You RASCAL. What are you thinking for gulp? just minification and shit or split into pieces and concat? I'm exploring a concat on my end and some qunit testing.

simeydotme:
Firstly just minification, gulp is a bit more hip/modern - and very easy to install 

judging by my small jaunt through the code; a concat would be kind of confusing unless there's some clear context changes in chunks of code in the upcoming releases - possible to separate `.prototype`s and `$.fn`s I suppose, but not really beneficial?

Issue #451, opened 2014-08-16
RTL support in combination with useCSS: false


andi1984:
For me RTL support in combination with useCSS: false doesn't work well for me in modern browsers (Chrome, FF etc.). On a one-step animation an image is shifted by multiple step-positions. It's spinning horizontally around. The one-step animation is definitely broken in combination with useCSS: false.

Enabling useCSS everything works as expected.

Would be nice if you could investigate this combination.

kenwheeler:
any news @ahmadalfy

ahmadalfy:
Found a fix for it, `_.currentLeft` should be set to its opposite value `-(_.currentLeft)` in case of RTL.
@kenwheeler would you like me to make a fiddle with it or submit a pull request directly?

kenwheeler:
Test and PR it, I'll focus on the rest of the confirmed bugs for 1.4

ahmadalfy:
URL for the test: http://alfy.me/slick/
Modification : http://alfy.me/slick/slick.js line 274(// Issue 451)
PR is coming shortly

kenwheeler:
#Woo

ahmadalfy:
#BINGO

ali-hellani:
great thanks, It works like a charm.

PR #952, opened 2015-01-24
Fixed issue #451, problem when using RTL with useCSS:false

ahmadalfy:


ahmadalfy:
The JS minifier I used generated a totally different file than your minified version. Can you take care of that?

kenwheeler:
Of course I can. Thanks for being the man and hooking it up

ahmadalfy:
Thanks, sorry for the delayed response. I am looking into the other RTL issues now as well

PR #966, opened 2015-01-29
Fix 'swipeToSlide' bugs, plus vertical slide variable height & new option 'adaptiveHeightMinVert'

digitor:
1. When 'swipeToSlide' option is true you can't slide backwards and if you have an odd number of slides, the last slide is not reachable.
Fiddle of fix below.
http://jsfiddle.net/87qm8zaz/14/

2. Changed so vertical slides to get the height of each slide, rather than assuming every slide is the same height.
Fiddle of feature below:
http://jsfiddle.net/87qm8zaz/13/

3. Added option 'adaptiveHeightMinVert' {number}, which accepts a minimum pixel height for the slide when in vertical mode. Fiddle is same as point 2 above.


jimdoyle82:
I'm on a PC, so couldn't build the make file. So I've deleted the min version to avoid confusion. I've leave that up to you. Cheers.

simeydotme:
Apologies @digitor that it took so long to get around to this.
But for now there would be far too much regression for me to entertain this PR... If you still think these problems exist, then I would happily re-evaluate a new PR after rebasing, and providing new Fiddles based on the current `master` branch.

Best, Simon.

Issue #970, opened 2015-01-29
Init Event


kenwheeler:
Init Event isn't firing

MariusRumpf:
The same seems to be the case for the `reInit` event. I would expect it to fire if one of the breakpoints in the `responsive` setting is reached. That isn't the case for 1.4.0.

kenwheeler:
the event must be bound prior to slick being called. not a bug

PR #979, opened 2015-01-31
Note about event listener before initialization

MariusRumpf:
For the init and reInit event to work correct the event listeners have to be added before initialization of slick. Added a note to the readme. #970


kenwheeler:
Good call!

PR #987, opened 2015-02-03
No need to do this

psquickitjayant:
Causing delay in the index change when slides are changed frequently using external control.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -320,12 +320,7 @@
                 _.$slideTrack.css(animProps);
 
                 if (callback) {
-                    setTimeout(function() {
-
-                        _.disableTransition();
-
-                        callback.call();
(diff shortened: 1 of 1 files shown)


kenwheeler:
This removes the callback wait when things are CSS animated. If you are using slickGoTo, use the dontAnimate argument

psquickitjayant:
I need to animate slides when using slickGoTo. Here it is my sample code. 
Using frequent switches from one slide to another by slickGoTo function is giving wrong index. 

```
$('.cn1inner-slider').slickGoTo(index, false);
```

kenwheeler:
If that works for your use case, that's cool, but it's not grounds to neuter the core library.

psquickitjayant:
Cool! Thanks!

PR #992, opened 2015-02-04
Add play/pause events

teohhanhui:


simeydotme:
Apologies for the delay.
I'm not sure this is useful, and there's some problems with autoplay already.
So we'll look to sort out these events in teh upcoming 2.0.

Best, Simon.

PR #1023, opened 2015-02-12
add a missing event "edge" in desktop version

arffak:
add a missing event "edge" in desktop version
http://jsfiddle.net/w839zzue/1/

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1747,6 +1747,10 @@
 
         _.currentLeft = _.swipeLeft === null ? slideLeft : _.swipeLeft;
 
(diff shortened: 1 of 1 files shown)


hbourquo:
Hi, thanks for this fix. However, I think the 'direction' event parameter is not correct in that case. It contains contains 'vertical' all the tme. Check your jsfiddle.

Thanks!

simeydotme:
This would be a nightmare in regression for `asNavFor` ... I also note it's causing multiple events being triggered erroneously.

I like the idea, but please consider creating a more rigid jsFiddle (specifically with `asNavFor` cases) and also fix the errors I've seen.

Thanks, Simon.

PR #1029, opened 2015-02-13
Add support for Composer

derhasi:
For future releases it was great you could add this composer.json and add the package to packagist.org :)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -36,6 +36,9 @@ bower install --save slick.js
 
 //NPM
 npm install slick-carousel
+
+//Composer
+composer require kenwheeler/slick-carousel
 ````
 
 #### Contributing
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,25 @@
+{
+    "name": "kenwheeler/slick-carousel",
+    "description": "the last carousel you'll ever need",
+    "type": "component",
+    "keywords": [
+        "responsive",
+        "carousel",
+        "jquery"
+    ],
+    "homepage": "https://github.com/kenwheeler/slick",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Ken Wheeler",
+            "email": "ken_wheeler@me.com"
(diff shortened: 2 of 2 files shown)


deefour:
I'm genuinely curious why this would be desirable. [Composer](https://getcomposer.org/) is the **"Dependency Manager for PHP"**, not Javascript. That's what [NPM](https://www.npmjs.com/), [Jam](http://jamjs.org/), [Bower](http://bower.io/), and others are for.

Would you suggest adding a `slick.gemspec` file too, so it can be added to RubyGems to be installable via [Bundler](http://bundler.io/)?

derhasi:
Composer is a dependency manager for PHP projects. That does not mean that the package have to be php, as you will use non-php packages in php projects too.

Besides that, npm, jam and bower are package managers that run on node. Why shouldn't a PHP based package manager not be allowed to handle Javascript-Packages?

I personally do not use Ruby for webdevelopment, but I do not see a reason, why that should not be a valid request.

If the maintainer of this project does not want that, that's fine. But then we would need to maintain a fork to do it anyways.

MariusRumpf:
Why not use the [composer-asset-plugin](https://github.com/francoispluchino/composer-asset-plugin) for this. This has the possibility to add sources from npm/bower in your `composer.json`.

derhasi:
Because it Is slow and it does only work when the plugin is installed globally. These are two crucial problems.

simeydotme:
I'm kind of on @deefour 's level here...
I've not seen a composer.json file in pretty much any frontend libraries before... https://github.com/jquery/jquery jQuery doesn't have it, so dependencies wouldn't be met, anyway?

I'll close this, but I'm pinging @kenwheeler so he can have one final [goosey](http://www.urbandictionary.com/define.php?term=goosey) and bring it back from the grave if he wants *composer.json* support :smile: 

Thanks for your effort, @derhasi . :)

PR #1045, opened 2015-02-17
Changed arrows and dot to use ISO entity code

tanmancan:
In certain cases, SASS was not compiling the prev-character, next-character, dot-character correctly.
May be a related issue: https://github.com/sass/sass/issues/1395

$slick-prev-character: "←" !default; The CSS output would turn into:
content: "ΓåÉ";

$slick-next-character: "→" !default; The CSS output would turn into:
content: "ΓåÆ"; 

$slick-dot-character: "•" !default; The CSS output would turn into:
content: "ΓÇó";

Using the ISO entities appears to have fix the issue, outputting the correct character in the CSS file:
$slick-prev-character: "\2190" !default;
$slick-next-character: "\2192" !default;
$slick-dot-character: "\2022" !default;

Into:
content: "←";
content: "→";
content: "•";


tanmancan:
Here is the jsfiddle for the merge:
http://jsfiddle.net/cbdznqgx/4/

simeydotme:
thanks @tanmancan ... can you please add comments to the sass file with the original ascii chars, so users reading it can identify the codes easily?

I'll merge pending that commit.

Best, Si.

tanmancan:
@simeydotme Added comments to show ascii output:
https://github.com/tanmancan/slick/commit/c902d6171371ddf5e75db6e46fe7bc089911eb96

PR #1051, opened 2015-02-18
Remove getNavigableIndexes counter assignment, which caused Chrome to crash when setting slidesToShow option dynamically

dylanfisher:
This pull request fixes an issue that causes Chrome to crash when dynamically setting the slidesToShow and slidesToScroll options dynamically. To reproduce the problem in the fiddle, resize your browser to the smallest breakpoint, then increase the window size. Chrome should crash.

http://jsfiddle.net/7uzkbeoj/2/

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -904,7 +904,6 @@
             if (_.options.centerMode === true) max = _.slideCount;
         } else {
             breakPoint = _.slideCount * -1;
-            counter = _.slideCount * -1;
             max = _.slideCount * 2;
         }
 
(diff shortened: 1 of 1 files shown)


kenwheeler:
What impact does this have on existing functionality? Wouldn't the counter be a necessary part of determining whether indexes are navigable?

kenwheeler:
It looks like your problem is that you aren't casting your slidesToShow to Int. You are providing a string. You should use parseInt and you won't have this problem.

dylanfisher:
Ahh you are right, the data-attribute is a string not an Int. I'll close this out, thanks for helping in any case!

PR #1072, opened 2015-02-24
Fix issue when using customPaging

binwilly:
This fix when using customPaging and change the html structure.

Problem: when you click in the child element of the <li>  the plugin look for the parent of the even.target and try to get the index.
Solution: Could use the event.currentTarget but for fallback compatibility ( < IE 8 ), I changed to look for the parents('.dot')

Here is the example:
http://jsfiddle.net/binwilly/sp2c342f/
If you click in the blue zone works as expected because is the <li> element, but in the <p> (red) always go to the first slide.


simeydotme:
Think this was fixed already.
Apologies.

PR #1086, opened 2015-02-27
Fix: pauseOnHover re-starts a paused carousel

joeholdcroft:
I needed a way to pause my Slick carousel temporarily. I have `pauseOnHover` set on the carousel. To pause the carousel, I run something like this:

```js
			$(this)
				.slick('slickPause')
				.slick('slickSetOption', 'pauseOnHover', false);

```

However, because the events are registered when the plugin gets initialised, they don't respect changes to the settings. As such, even after running this code I can re-start the carousel by hovering on it then moving my mouse elsewhere.

To resolve this, I moved the if statements inside the event listeners.

This is probably an issue for a handful of other settings, too.

JSFiddle showing the bug: http://jsfiddle.net/15h172h1/

Same JSFiddle with my forked version of Slick: http://jsfiddle.net/a9pm4oL5/1/

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1043,26 +1043,24 @@
 
         _.$list.on('click.slick', _.clickHandler);
 
-        if (_.options.autoplay === true) {
-
-            $(document).on(_.visibilityChange, function(){
+        $(document).on(_.visibilityChange, function(){
+            if (_.options.autoplay === true) {
                 _.visibility();
-            });
-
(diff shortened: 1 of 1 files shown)


simeydotme:
exemplary issue / pull request ! :+1:

joeholdcroft:
@simeydotme great! Anything I need to do to get this merged?

kenwheeler:
Thanks @joeholdcroft

PR #1096, opened 2015-03-04
add infiniteAutoplay option

wuxiaoyi:
If infinite is false, I would meet a problem when I apply autoplay. I want to set infinite as false is because it will stop when Carousel at the last page. But now, when the Carousel at the last page, it continue to display from end to front. I think it goes against to meaning of infinite. Thus, I delete the code that will change direction  by autoPlayIterator in my project. In the code of I commit to slick, to be safe, I add an infiniteAutoplay configuration that if the value is true, it will change direction, if the value is false, Carousel will stop.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -65,6 +65,7 @@
                 fade: false,
                 focusOnSelect: false,
                 infinite: true,
+                infiniteAutoplay: true,
                 initialSlide: 0,
                 lazyLoad: 'ondemand',
                 mobileFirst: false,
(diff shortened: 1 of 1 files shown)


simeydotme:
http://jsfiddle.net/e73fvm3o/
^^ here's a JSFiddle for you.

No idea what you're trying to achieve :) - looks like you want to stop at the last slide? - this defeats the idea of autoplay? if you want to stop autoplay on the last slide you can do this yourself in the `onAfterChange` event ...

simeydotme:
```js
$(".s3").slick({
    infinite: false,
    autoplay: true,
    autoplaySpeed: 1000
}).on("afterChange", function(e, slick, cur) {
      
    if( cur === slick.$slides.length-1 ) {
        slick.setOption("autoplay", false, true);
    }
    
});
```

Issue #1069, opened 2015-02-23
Dots navigation does not work when slidesToShow and slidesToScroll are equal and the number of slides does not divide into slidesToShow evenly


choppsta:
You can recreate the issue on the demo page in the "Add & remove" example.  The example has slidesToShow = 3 and slidesToScroll = 3.  If you add 6 or 9 slides the dots work fine.  If you have 5 or 8 slides the dots no longer work.

I've tracked down the source of the problem to Slick.prototype.getNavigableIndexes() but I've not been able to work out why it's not working.

yesdevnull:
After a lot of playing around I've managed to get it working.  Your hunch about the issue being in [Slick.prototype.getNavigableIndexes()](https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L898) was bang on!

### tl;dr quick fix
```js
counter = (_.slideCount * -1) -1; // Line 907
```

### Details
I have a carousel with 14 items, and 3 items per slide.  When clicking the nav dots, they would always be one dot behind.  I went into [slideHandler](https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L1719) and put in `console.log('slideHandler: ' + index);`.  In the Console, I'd see:

```
slideHandler index: 0
slideHandler index: 3
slideHandler index: 6
```

And so on and so forth, you get the picture.  However, a `console.log(navigables)` (on [Line 641](https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L641)) gets this result:

```
[-14, -11, -8, -5, -2, 1, 4, 7, 10, 13, 16, 19, 22, 25]
```

That doesn't look right, the indexes are off by 1.  The function that populates `navigables` is [Slick.prototype.getNavigableIndexes](https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L898).  

The fix for me was to change `counter` to have -1 more than the total slides.  For example, `14 * -1` is `-14`, which would make the dots click 1 index behind.  By changing the calculation to `(14 * -1) -1`, you now get `-15`.  Now, my math skills aren't particularly strong, but I think the issue is because arrays start at 0.

```js
counter = (_.slideCount * -1) -1; // Line 907
```

Now, a log of the `navigables` array shows:

```
[-14, -12, -9, -6, -3, 0, 3, 6, 9, 12, 15, 18, 21, 24, 27]
```

[...]

Issue #1039, opened 2015-02-16
Pager incorrect with Show Multiple / Scroll Multiple


jahshuwaa:
I noticed two problems with the pager on a site I'm building but it can more easily be seen on http://kenwheeler.github.io/slick/

1. Go to Section: "Add & Remove"
2. Add 12 items, to have a total of 13 items
3. Click the last (5th) pager
4. Click the 1st Pager... notice it scrolled to "11 12 13"
5. Keep clicking the 1st Pager. It re-scrolls to "11 12 13"

Another issue with the pager, also seen here, is that the nesting of elements inside the pager LI seems to be causing an incorrect "index" broadcast. For example, making the following tweaks to the CSS (using Chrome inspector):

```css
.slick-dots li {
width: 50px;
height: 50px;
background-color: #f00;
}
.slick-dots li button {
padding: 0;
background: #ccc;
}
```
After this tweak, clicking the red area has strange consequences.

It gets worse if another element is nested inside the pager, for example if using a customPager such as this, with different padding around the span and div:
```javascript
customPaging: function(slider, i) { 
return '<span><div></div></span>';
}
```
If the slick.click is fired from the div vs. the span, results are strange.

I have these issues on a site near launch so I will be moving to a different carousel platform. Just thought I'd present here since it's also on the demo site.

![2015-02-16_10-43-24](https://cloud.githubusercontent.com/assets/784825/6214845/a7d34eee-b5c9-11e4-9158-68d1c10d2ee6.png)

dweerd:
A codepen for the first issue: http://codepen.io/anon/pen/zxjzNx

yesdevnull:
My answer on Issue #1069 has a fix that I hope fixes the issue for you!

PR #1104, opened 2015-03-06
[BUGFIX] Navigation dots not navigating correctly

yesdevnull:
Building off #1069 and #1039 there's an issue with the navigation dots with certain numbers of slides and `slidesToScroll`.

For example, if you have 14 slides and have `slidesToScroll`/`slidesToShow` set to 3, then start clicking the navigation dots, you'll generally be presented with the 3 previous slides (and missing 1 slide, usually the first.)  I have set up 2 fiddles, one that demonstrates the problem, and another fiddle that has the exact same HTML and JavaScript, just with my fixed version of Slick.

[JSFiddle that doesn't work correctly](http://jsfiddle.net/094obpe5/1/)
[JSFiddle that is fixed](http://jsfiddle.net/3pgksu48/6/)

This fix resolves the issue where either `slideCount` or `slidesToScroll` is odd, and `slidesToScroll` isn't a factor of `slideCount`.  Having said that, this can't resolve the issue of the offset of `slideCount` and `slidesToScroll` being greater than 1.

I hope I've described this issue correctly, and that you fix the patch appropriate.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -904,7 +904,11 @@
             if (_.options.centerMode === true) max = _.slideCount;
         } else {
             breakPoint = _.slideCount * -1;
-            counter = (_.slideCount * -1) -1;
+            if (((_.slideCount % 2) == 1) || ((_.options.slidesToScroll % 2) == 1)) {
+                counter = (_.slideCount * -1) -1;
+            } else {
+                counter = _.slideCount * -1;
+            }
             max = _.slideCount * 2;
         }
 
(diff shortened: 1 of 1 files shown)


d3x42:
line 902. Infinity=false and bigger max work for me: ` if(_.options.infinite === false) {
            max = _.slideCount;`

john-landgrave:
@yesdevnull Thanks for the hard work and digging on this one! I did end up needing to use @bpander's fix on what is now line 1181 (after applying your code). This was on a carousel with 4 visible, scroll by 4, 13 items.

PR #1110, opened 2015-03-08
add pause requested variable that doesn't switch off on hover

fredkingham:
hey Ken,

so my problem was when you explicitly pause the carousel, a mouse over will then restart it (if pause on hover is set to true)

here's the problem in action
http://jsfiddle.net/d7ecuaxw/2/

this pull request fixes it as you can see
http://jsfiddle.net/n5cqfrtq/4/

Let me know what you think, will gladly alter it if you think its necessary.

Cheers,

 Fred

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -123,6 +123,7 @@
             _.cssTransitions = false;
             _.hidden = "hidden";
             _.paused = false;
+            _.pauseRequested = false;
             _.positionProp = null;
             _.respondTo = null;
             _.shouldClick = true;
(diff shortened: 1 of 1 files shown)


simeydotme:
Is this related?
https://github.com/kenwheeler/slick/pull/1086

fredkingham:
ahhh yes it is, and Joe's fix is better, closing.

fredkingham:
actually scratch that, it looks like that bug fix needs to applied to the initDotEvents as well, so the bug still exists.

Its only been fixed in the event that the dots aren't there. I still like Joe's fix better, but I'll create a new pull request later, to cover the dots

simeydotme:
:+1:

fredkingham:
btw, pull request submitted #1114

PR #1112, opened 2015-03-09
Wrote tests for Slick

yesdevnull:
As there are currently no tests and easy way to ensure any changes you make to Slick don't break it, I've made tests that are copies of the demos on the [Slick site](http://kenwheeler.github.io/slick/).  This way, you can make your required changes to Slick then test the modes between the Slick site and your tests to ensure you have fixed your problem, and haven't introduced any issues.

All tests use the same code as the examples on the Slick website.  The reason for splitting each demo into a separate file is to avoid any conflict/confusion from any other JavaScript running on the page.  I would consider each file a "clean room" for testing.

### DIFF
--- a/tests/01-single-item.html
+++ b/tests/01-single-item.html
@@ -1,0 +1,36 @@
+<!doctype html>
+	<html lang="en">
+		<head>
+			<link href="../slick/slick.css" rel="stylesheet" />
+			<link href="../slick/slick-theme.css" rel="stylesheet" />
+			<link href="style.css" rel="stylesheet" />
+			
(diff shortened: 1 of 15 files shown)


kenwheeler:
Hey I'm not gonna pull this in right away because I want to add some automated test running to it. Thanks man!

yesdevnull:
Sure thing.  Need any assistance with writing tests?  Happy to provide a hand in getting them finished.

simeydotme:
@kenwheeler is this still valid considering your own testing setup?

PR #1114, opened 2015-03-09
extending Joe's fix to cover pause on hover

fredkingham:
so we also needed to fix the pause on hover on dots

extending #1086 

this is broken
http://jsfiddle.net/k6jq2ym6/8/

this is fixed
http://jsfiddle.net/k6jq2ym6/6/

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1006,19 +1006,21 @@
             }, _.changeSlide);
         }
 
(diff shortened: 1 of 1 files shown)


simeydotme:
should be fixed on master, thanks for your effort.
apologies for the delay.

PR #1151, opened 2015-03-17
Adaptive height enabled for multiple visible slides

SimonShirley:
Adaptive height is now calculated from all the slides shown. The largest
height is then set as the height for the $list.


hakastein:
Why this pull request isn't applied?

simeydotme:
because it has no tests, or examples to show it working, or why it was needed.
if the author doesn't do it, then it can take a long time for a contributor to do it, instead.

it also doesn't look like a core/needed feature, so actually I will close it and put it under consideration for 2.0.

SimonShirley:
Apologies @simeydotme, I didn't realise this was needed.

Please find a CodePen of Slick in its current release (1.5.9): https://codepen.io/simonmshirley/pen/remoXb

When adaptive height is enabled, and slidesToShow is set to 1, the height of the slider adapts to the height of the visible slide, as expected.

When multiple slides are visible, the adaptive height stops working and the height of the slide is fixed to the largest slide out of the collection:
https://codepen.io/simonmshirley/pen/vGmbJJ

This change (with the comment change - https://github.com/SimonShirley/slick/commit/19ec104cea9cbd904259edc4873d71f2f23d0b59 - although the commit version works too) now calculates the height from the visible slides, as shown in the following CodePen: https://codepen.io/simonmshirley/pen/yObZEm

simeydotme:
Thanks @SimonShirley ... actually I just took a look through your code... and without even having to run it, I can see it's not going to work with `centerMode` or other edge cases to do with clones ... I think unless you have a lot of time to spend to update this and run it through a bunch of setting configurations... then we'll just have to leave it closed :disappointed: 

Do really appreciate your effort, and we'll do our best to get it sorted in the future if you don't have the time.

Best, Si.

PR #1159, opened 2015-03-19
Updated README.md to reflect slickGoTo syntax

jlyman:
The documentation was missing the second parameter for slickGoTo, added it in. Also, the marketing site has confusing documentation for that second parameter (I thought setting it to true would cause it to animate, not skip the animation). 

Thanks for the fantastic library!


kenwheeler:
Thanks dude!

PR #1160, opened 2015-03-19
Chcange possition of navigation arrows

Igloczek:
After this change we can use absolute positioning on arrows using simple CSS - `top:0; bottom: 0; margin: auto;` (without negative margins), when outer container is higher than slider content (i.e. with dots positioned relativley).


kenwheeler:
Wait, why not just pass the list to the appendArrows option?

Igloczek:
Ah, true.
BTW. Thanks for great plugin.

PR #1184, opened 2015-03-26
fixed .unload() and .destroy() for jquery string selectors

Keyes:
currently it isn't possible to use .destroy() or .unload() when using jquery selector strings for the arrows. If you do so, your arrows simply will be removed, since they are obviously not an object or DOM node. This fixes this wrong behaviour.


simeydotme:
I belive this is no longer an issue :)

PR #1192, opened 2015-03-30
Improve add remove methods

Ninigi:
### After/Before Add/Remove events
*slickAdd, slickRemove* triggers before and after events, passes slick and removed/added element

### Remove Element by passing jQuery Object or selector string
Since I struggled to get the right index for the object I wanted to remove, I added this feature. Simply pass the jQuery collection, or a selector String. If the collection includes more than one object, it will remove all objects in the collection.

[fiddle it out](http://jsfiddle.net/1rdjt203/8/)


kenwheeler:
Awesome, let me review this a bit further

simeydotme:
this looks really good...
any chance you can rebase @Ninigi so I can review and merge? :)
Apologies for the delay.

nikowuesti:
I implemented this in my own copy and it works quite well, thanks for that!

Issue #999, opened 2015-02-05
Detroying Slick (Unslick) Removes External Events


arpowers:
hey there, 
Using slick with Knockout, when i add/edit new elements I prefer to just reset the carousel rather than deal with the logic of adding them via individual functions. 

Anyway, the 'unslick' command also removes bindings associated with Knockout rendering this technique impossible in my case.

Ap

simeydotme:
Hey, thanks for contributing @arpowers  :)

Of course removing slick deletes the slick nodes and puts your code back to pre-slick state... so the DOM nodes you were binding to are gone. -- perhaps call your knockout data-binds from a function and call that function on slick-init? (I have extremely limited knowledge of knockout.) :smile: 

This question is better suited for www.stackoverflow.com where there are thousands of people with knockout.js knowledge and can help you with your code theory/architecture... this doesn't seem like a slick issue/bug. :smile: 

If you really don't want to go there, create a JSFiddle as asked for in the https://github.com/kenwheeler/slick/blob/master/CONTRIBUTING.markdown and set it up in your example and we might be able to help.

Simon.

arpowers:
Well, aren't you just supposed to namespace your events and return to the same state as when you start? 

I got around it using .clone() so I'm ok either way

kenwheeler:
Slick events are namespaced. Without a JSFiddle I can't see what exactly is going on with Knockout.

nickmelville:
I think this might be a simplified example of the problem, but if not, I can open a separate issue: http://jsfiddle.net/woL708Lz/2/

1. Click a link (event target is logged in console).
2. Click 'slick' to make the slider.
3. Click a link (event target is still logged).
4. Click 'unslick' to destroy the slider.
5. Click a link and the event target is no longer logged in console.

hidanielle:
Am having this issue as well... #1199 fixes the addressed issue but it creates more problems and is unusable

Issue #1079, opened 2015-02-25
After calling "refresh" click events won't fire anymore


SunboX:
See Demo at: http://jsfiddle.net/SunboX/j1fvz4mh/4/

nickmelville:
The issue is actually with the destroy method.  Refresh essentially calls 'destroy' and then 'init'.

Updated example: http://jsfiddle.net/q8ejbaga/1/ (click "unslick" and the events stop working).

I believe this is duplicating #999.

webdesignberlin:
Same Problem. looks like all event handler will be deleted.
![Image Event Listener](http://s1.postimg.org/rf7bmsbm7/unslick_before_after.png)

itrupti:
I have declared click method as windows method and tried to call it from slide. I have updated nick's jsfiddle. This is just workaround. 

http://jsfiddle.net/itrupti/woL708Lz/3/

simeydotme:
This appears to be working with the latest version.

PR #1199, opened 2015-04-02
Bugfix #999 and #1079 (Unslick destroyed EventHandler)

webdesignberlin:
Replace .html() with .prepend() on destroy function, cause: When .html() is used to set an element's content, any content that was in that element is completely replaced by the new content. Additionally, jQuery removes other constructs such as data and event handlers from child elements before replacing those elements with the new content. [https://api.jquery.com/html/#html2](Find on jQuery API). Fixed Issue #999 and #1079.

[http://jsfiddle.net/webdesignberlin/ohnmp0yu/5/](original error) now [fixed](http://jsfiddle.net/webdesignberlin/ersybgam/1/)

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -757,7 +757,7 @@
                     width: ''
                 });
 
-            _.$slider.html(_.$slides);
+            _.$slider.prepend(_.$slides);
         }
 
         _.$slider.removeClass('slick-slider');


kenwheeler:
Thanks for this, will review fully soon

itrupti:
Hi,

I tried solution suggested by webdesignberlin but now i am facing another issue.

As suggested it prepends slides to existing elements so number of dots and number of slides calculation comes wrong.

Please see the html structure in the attached screenshot.

slick issue


Please let me know there is any workaround for same.

sumanita:
hi @webdesignberlin i am using latest version of carousel and looks like this issue reappears as it remove my event handler on orientation change. any help will be highly appreciated. 
thanks in advance :)

adjustmentlayer:
slider - element on which slick was initialized;
after destroy - event which slick fires after it has destroyed slider
expander - plugin that takes text and cuts it into two parts. Last part is hidden and you can expand it. 
To do that this plugin adds extra markup. Thats all you need to know to understand problem. https://kswedberg.github.io/jquery-expander/
destroy - `$('.some-el').slick('unslick')`

I just reset all handlers inside 'slider' 'after destroy'. 

But when i used this plugin inside slider slides 
I ran in problem that after slider destroy. I can not destroy 'expander'. 
It is important because previously expander added extra markup. 
And if i just reinit expander it will wrap this extra markup in another extra markup.

Of course I can destroy expander before 'destroy'
but if i call 'destroy' in multiple places in my app. My code would repeat

And I think slick needs to fire event not only after 'destroy' but also before 'destroy'. 
Currently such features is absent;

Issue #1088, opened 2015-02-27
When total slides not in multiple of slidesToShow, clicking last dot doesn't go to last slider page


analuisamartins:
Hi,
I found this issue when scroll infinite = false, dots = true and total slides is not multiple of slidesToShow. 
(Ex: 10 slides, 4 per page) Click in last dot never triggers last page.
After digging a little bit the code, I found that the problem may be in the method Slick.prototype.getNavigableIndexes = function() {... 
I replaced the calculation of max by the following:
        if(_.options.infinite === false) {
            // max = _.slideCount - _.options.slidesToShow + 1;
            max = Math.ceil(_.slideCount / _.options.slidesToShow) * _.options.slidesToShow;

It seems to work fine now, but as I don't know very well the plugin, I don't know if is the right fix (can cause collateral bugs).

Great plugin btw ;)

shabith:
yeah this did the trick for me as well. @analuisamartins thanks for sharing this. :+1:

marcinkrzeminski:
Had the same problem. Thanks @analuisamartins for quick fix ;).

carasmo:
When will this be corrected in the repository?

BTW: Thank you for the fix, it works like it did in 1.3.x, which was correc

carasmo:
Thank you for this slider too, didn't mean to sound ungrateful. Just spent hours trying figure out why it worked in 1.3 and not worked in 1.5, so slightly grumpy, but whatever :-)

Cottser:
Thanks @analuisamartins, this was almost a show stopping issue on a project I'm working on. @kenwheeler would a PR help in moving this forward? JSFiddle/JSBin/etc. test case?

kenwheeler:
All of the above would be helpful

Cottser:
Okay, I'm on it.

PR #1239, opened 2015-04-15
Fix for last dot not working in some cases in infinite mode

Cottser:
More details are in #1088.

For the test cases, clicking the second dot is how I've been testing.

Before: http://jsfiddle.net/spgx6nzh/
After: http://jsfiddle.net/maad9poa/3/

I'd be happy to expand the fiddles to include more scenarios but might need a bit of guidance.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1006,7 +1006,7 @@
             max;
 
         if (_.options.infinite === false) {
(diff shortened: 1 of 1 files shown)


Cottser:
This has been fixed elsewhere (the "Before" fiddle works now), closing.

Issue #1204, opened 2015-04-05
custom pagination not work 1.5


cansozeri:
After updated to 1.5 my custom paging is not working anymore. What has changed ?? 

It is working with 1.4.

Here is my settings.

```
$('#slider_main').slick({
    slide: '.slider_img',
    arrows: false,
    dots: true,
    autoplay: true,
    autoplaySpeed: 5000,
    appendDots: '#slider_main .slider_pagination',
    customPaging: function(slider, i) {
        return '<a href="javascript:void(0)">' + (i + 1) + '</a>';
    }
});
```

You can see here http://jsfiddle.net/2mmcqc2m/13/

pagination div should not be loaded as a slide.... Because it has not got "slider_img" class..

I have found it, I don't know why but on this function ".options.slide" has removed, so slide option is not working. 

```
//current one

_.$slides = _.$slider.children(
            ':not(.slick-cloned)').addClass(
            'slick-slide');

//It should be

 _.$slides = _.$slider.children(_.options.slide +
            ':not(.slick-cloned)').addClass(
            'slick-slide');
```

leggomuhgreggo:
Looks like this is out of date, and I think it's been fixed. Apologies for the dormancy. Thanks

Issue #293, opened 2014-06-10
CustomPaging doesnt work with html snippet


alaknanda:
I tried something like this :
<pre>
customPaging: function(slider, i) {
                  return '&lt;div&gt;&lt;p&gt;abcd&lt;/p&gt;&lt;p&gt;qwerty&lt;/p&gt;&lt;/div&gt;'
                },
</pre>
The slides dont change .
<pre>
 However, if the html is changed to &lt;div&gt;abcd&lt;/div&gt;, then the pagination works fine.
</pre>

aaronstezycki:
Not sure if this has been fixed yet? Just gonna bump this so it gets noticed again. @kenwheeler

gigi:
String 667
Slick 1.5.0
I think it has to be changed

event.data.index || $(event.target).parent().index() * _.options.slidesToScroll;
to
event.data.index || $(event.target).index() * _.options.slidesToScroll;

simeydotme:
Recently fixed by: https://github.com/kenwheeler/slick/commit/f66ac9e155f249f8d034a81ed61793c6b56efbd7 :kissing_heart:

daysable:
@simeydotme, thank you for that jsfiddle example. But how would add active states to the custom pagination? I'm having difficulty there.

daysable:
Nevermind! I figured it out.

rijosh:
@simeydotme  How can i append this custom pagination to a particular div. cause i need to one button near this pagination 
![image](https://user-images.githubusercontent.com/9720488/30023484-b189fa38-9180-11e7-805f-9856691e0b12.png)

PR #1251, opened 2015-04-19
Fixes issue with customPaging children click event

jamiehs:
Test case fiddle: http://jsfiddle.net/jamie3d/tah9tndm/1/

Not that clicking the "Foo" text (appended directly) works as expected. Clicking any of the other child elements fails to fire due to the indicated change.

Similar to #1204 and #293

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -661,7 +661,7 @@
 
             case 'index':
                 var index = event.data.index === 0 ? 0 :
-                    event.data.index || $(event.target).parent().index() * _.options.slidesToScroll;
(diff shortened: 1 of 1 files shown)


cansozeri:
The problem is still occurs and this fix is not working with 1.5.5

simeydotme:
This was fixed in `1.5.3` with a more robust solution :)
https://github.com/kenwheeler/slick/commit/f66ac9e155f249f8d034a81ed61793c6b56efbd7

simeydotme:
@cansozeri you're talking about something else, this issue is about customPaging (dots) not slides. :smile:

cansozeri:
@simeydotme ooops sorry : )

PR #839, opened 2014-11-28
Add reinit on specific slide

keithlam:
Allow reinit of Slick to specify a current slide to address the issue when a slide is inserted at the beginning of the carousel dynamically on "previous" click.  This allows lazy-loading of very long carousels.
https://github.com/kenwheeler/slick/issues/303

PR #1262, opened 2015-04-22
Reinit on Specific-Slide

hereandnow:
This pull request is addressing issue #839. in the pull-request from @keithlam the Call to `_.setSlideClasses` had always `0`as an argument, which was wrong (the class `slick-active` was not reinitialised then). 

I added that one. Not that sure if this is everything which needs to be done, but i think so.


keithlam:
Thanks for reviving the fix for this issue!  But I think there was a change to slick.js itself.  That call had a different check instead of just hard-coded 0.
https://github.com/keithlam/slick/blob/issue-303/slick/slick.js#L494

simeydotme:
think this was resolved.

PR #1275, opened 2015-04-24
Add option `dotsMax` for keeping number of dots under a maximum

LoicMahieu:
http://jsfiddle.net/mahieuloic/ubmgbwj1/

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -72,6 +72,7 @@ cssEase | string |  'ease' | CSS3 easing
 customPaging | function | n/a | Custom paging templates. See source for use example.
 dots | boolean | false | Current slide indicator dots
 dotsClass | string | 'slick-dots' | Class for slide indicator dots container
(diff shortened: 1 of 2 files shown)


simeydotme:
Hey, Yo @LoicMahieu ... Thanks for contributing...
Could you please add a little explanation, use-case example and such with the reason behind the pull request? is it associated with an issue? :)

Regards, Simon.

LoicMahieu:
Hi @simeydotme 

Thanks for your response. Your fiddle is exactly what I wanted, a slider with a lot of items. I didn't think to use `init` event.
No I did not open a issue for that.

The best would be to determine how many dots can be displayed but it's far more complicated.
As far as you can implement with "manually", this PR is not so useful... So closed it if you want.

simeydotme:
:)
You could also play around with the CSS and `overflow: hidden;`if you wanted to only show the amount of dots that fit your display.

PR #1279, opened 2015-04-27
Add 'destroy' event to slick

ducdigital:
This is quite useful. One of the use case is that I use Init to remove several class that affect dom of Slick. After unslick it should be added back to the dom.


simeydotme:
I had to make some amends, because we didn't want the `destroy` event firing on every breakpoint. :smiling_imp: - please look it over if you wish.

Si.

PR #1296, opened 2015-04-30
fixes the null pointer exception in set position when called from outside...

Muenze:
It's my first pull request, please be gentle in review ;)


Muenze:
It fixes the null pointer exception when setPosition is called multiple times from outside, propblem occured when using setPosition in a directive that was in every slide.

simeydotme:
Muenze, can you explain this a bit clearer and maybe demonstrate the problem you fixed in a JSFiddle? :smiley:

PR #1298, opened 2015-04-30
Allow 'asNavFor' parameter to be jQuery object.

limonte:
Allow this usage of `asNavFor` parameter: 
```js
$('.post').each(function() {
  var container = $(this);

  container.find('.slider-for').slick({
    asNavFor: container.find('.slider-nav')
  });

  container.find('.slider-nav').slick({
    asNavFor: container.find('.slider-for')
  });
});
```

PR #1305, opened 2015-05-02
AsNavFor, Breakpoint and Init Events.

### LINKED ISSUE #865: Add ability to use asNavFor for many slicks

simeydotme:
___JSFiddle: http://jsfiddle.net/0u5qyg8x/___

- Merge in and amend @limonte asNavFor patch - allowing use of `$()` objects as targets.
- Fix up a small bug in @limonte patch as it was incorrectly type-checking for `null`
- Add the ability to `asNavFor` on mulitple slick instances.
- Fix up `breakpoint` events so they only fire during `resize` and not on `init`
- Fix up `init` event so it doesn't fire on `resize` anymore.
- Add `unslick` event which will also expose the breakpoint at which it unslicked (if possible).  
- Take cue from @muddylemon [branch](https://github.com/muddylemon/slick/commit/6e44c4f9b1b6df214a27672074d35b33b7c3d42e#diff-eaaa72f2e5f1d4974e4f1ae902717908R340) to prevent using `self` as `asNavFor` target.

===
- closes #865
- closes #1298 
- closes #1150 

===
[This patch's jsfiddle](http://jsfiddle.net/0u5qyg8x) also neatly demonstrates some of the bugs in `Slick.prototype.slideHandler()` regarding non-infinite-carousels as targets from infinite-carousels (wrong index is passed). I had a crack at that, too... but it's pretty damn crazy up in that method! So left it alone for now.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -645,6 +645,7 @@
                 }
             }
 
+            // only trigger breakpoints during an actual break. not on initialize.
             if( !initial && triggerBreakpoint !== false ) {
                 _.$slider.trigger('breakpoint', [_, triggerBreakpoint]);
             }
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -191,7 +191,7 @@
             // Extracted from jQuery v1.11 source
             _.htmlExpr = /^(?:\s*(<[\w\W]+>)[^>]*)$/;
 
-            _.init();
+            _.init(true);
 
             _.checkResponsive(true);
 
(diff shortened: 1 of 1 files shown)


simeydotme:
Going to Merge this, @kenwheeler - It's stable and well tested locally and on JSFiddle provided. But as mentioned; the fiddle highlights a few `slideHandler` bugs when used from `asNavFor`.

PR #1306, opened 2015-05-02
check if the asNavFor target has been unslicked, or not.

### LINKED ISSUE #1304: Unslicking in Responsive Breakpoint doesn't deactivate the adding of slick-active slick-center
Hello,

This is a branched issue from #1303 
That issue outlines the issue with reslicking after it's been unslicked by the responsive breakpoint.

This issue is related but seperate. If you unslick at certain breakpoints this works fine, however if you leave it for a little bit of time you'll find the individual elements start getting the 'slick-active slick-center' classes applied to them one at a time.  So it seems the unslick didn't cancel the timer for the addition of those classes.

Thank you

simeydotme:
Resolves #1304 

- __Before:__ http://jsfiddle.net/odoqefdq/
_(notice the unslicked carousel will get `.slick-active` and `aria-hidden=false` applied to it.)_

- __After:__ http://jsfiddle.net/odoqefdq/1/
_(now the unslicked carousel will not get attributes added, because `asNavFor()` doesn't run on it.)_

===

@kenwheeler could you review this one, please, because I added a kind-of architecture piece to it on line 852, regarding a `unslicked` state after a carousel has been destroyed. I couldn't see a valid and efficient way to check for an "_unslicked_" carousel in the exisiting code. :confounded:  -- You might have a better/different thought on how to do this. :smile:

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -347,7 +347,10 @@
 
         if ( asNavFor !== null && typeof asNavFor === "object" ) {
             asNavFor.each(function() {
-                $(this).slick('getSlick').slideHandler(index, true);
+                var target = $(this).slick('getSlick');
+                if(!target.options.unslicked) {
+                    target.slideHandler(index, true);
+                }
             });
         }
 
@@ -845,6 +848,8 @@
 
         _.$slider.removeClass('slick-slider');
         _.$slider.removeClass('slick-initialized');
+
+        _.options.unslicked = true;
 
     };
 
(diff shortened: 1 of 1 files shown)


slick/slick.js
```diff
@@ -846,6 +849,8 @@
         _.$slider.removeClass('slick-slider');
         _.$slider.removeClass('slick-initialized');
 
+        _.options.unslicked = true;
```
kenwheeler: This doesn't need to be on options, you can set it right on `_` like `animProp`. Its an internal instance variable, not a configurable option

simeydotme: aight, good point! - see, I knew code review was good for something!  :stuck_out_tongue_winking_eye: 
I'll sort it out after my :ramen: !

PR #1324, opened 2015-05-08
Added ajax-loader.gif to bower.json

leevigraham:


xer0x:
+1 because this makes using gulp & main-bower-files easier

PR #1327, opened 2015-05-08
Bugfix 1315

### LINKED ISSUE #1315: Multiple responsive sliders break when new dots event fired
I was able to replicate this with the latest version (1.5.2), 1.4.1 doesn't seem to have an issue with the bug though, where as 1.5.0 does.

If you have multiple sliders on a page which some by default have dots enabled and others don't, if a new breakpoint happens and dots are set to true where they previously weren't then it will kill other sliders dot interactions. One way to restore the behaviour back to the dots is to either use arrows or swipe action to move to the next pane, after this the behaviour is back until the next breakpoint.

I've setup an external demo to show, if you reduce the page width to where the bottom slider has dots appear you'll notice the top slider dot interaction stops working. Same happens if you have dots enabled by default and remove them at a later breakpoint, when you scale back up the dot event being fired breaks other image sliders.
http://jsfiddle.net/rr0zxoeL/

    $('.responsive').slick({
      dots: true,
      infinite: true,
      speed: 300,
      slidesToShow: 2,
      slidesToScroll: 2
    });

    $('.responsive2').slick({
      infinite: false,
      speed: 300,
      slidesToShow: 1,
      slidesToScroll: 1,
      responsive: [
        {
          breakpoint: 500, // this breakpoint kills .response dots
          settings: {
            slidesToShow: 1,
            slidesToScroll: 1,
            infinite: true,
            dots: true
          }
        }
      ]
    });

simeydotme:
Fixes #1315 and a small oversight in #1312

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -733,14 +733,18 @@
 
         var _ = this;
 
-        if (_.options.dots === true && _.slideCount > _.options.slidesToShow) {
-            $('li', _.$dots).off('click.slick', _.changeSlide);
-        }
+        if (_.options.dots && _.$dots !== null) {
 
-        if (_.options.dots === true && _.options.pauseOnDotsHover === true && _.options.autoplay === true) {
-            $('li', _.$dots)
-                .off('mouseenter.slick', $.proxy(_.setPaused, _, true))
-                .off('mouseleave.slick', $.proxy(_.setPaused, _, false));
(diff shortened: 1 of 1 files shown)


nbankzy:
Very similar issue is still present, when dots are changed from false to true, other slides break. Have commented the source on #1315

PR #1334, opened 2015-05-11
Bugfix 327 - Responsive Settings via setOption

### LINKED ISSUE #327: "Responsive" option isn't being updated in slickSetOption
I'm working some more on my Drupal module and was making an update function so that users can update a Slick instance in code if they need, and I noticed that even the most basic of responsive objects isn't working with slickSetOption. Might just be an oversight on my part, but I figured I'd post this here anyways. Below is an example of my code:

```
var key = 'responsive';
var fieldSettings = '[{
            breakpoint: 768,
            settings: {
              slidesToShow: 4,
              slidesToScroll: 4
            }
          },
          {
            breakpoint: 600,
            settings: {
              slidesToShow: 3,
              slidesToScroll: 3
            }
          },
          {
            breakpoint: 480,
            settings: {
              slidesToShow: 2,
              slidesToScroll: 2
            }
          }]';
$(target).slickSetOption(key, fieldSettings, true);
```

Any help would be greatly appreciated. Thanks!

simeydotme:
Fixes #327

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -35,8 +35,7 @@
 
         function Slick(element, settings) {
 
-            var _ = this,
-                dataSettings, responsiveSettings, breakpoint;
+            var _ = this, dataSettings;
 
             _.defaults = {
                 accessibility: true,
@@ -144,27 +143,6 @@
             _.currentSlide = _.options.initialSlide;
 
             _.originalSettings = _.options;
-            responsiveSettings = _.options.responsive || null;
-
-            if (responsiveSettings && responsiveSettings.length > -1) {
-                _.respondTo = _.options.respondTo || 'window';
-                for (breakpoint in responsiveSettings) {
-                    if (responsiveSettings.hasOwnProperty(breakpoint)) {
-                        _.breakpoints.push(responsiveSettings[
-                            breakpoint].breakpoint);
-                        _.breakpointSettings[responsiveSettings[
-                                breakpoint].breakpoint] =
-                            responsiveSettings[breakpoint].settings;
-                    }
-                }
-                _.breakpoints.sort(function(a, b) {
-                    if (_.options.mobileFirst === true) {
-                        return a - b;
-                    } else {
-                        return b - a;
-                    }
-                });
-            }
 
             if (typeof document.mozHidden !== 'undefined') {
                 _.hidden = 'mozHidden';
@@ -192,8 +170,9 @@
             // Extracted from jQuery v1.11 source
             _.htmlExpr = /^(?:\s*(<[\w\W]+>)[^>]*)$/;
 
-            _.init(true);
 
+            _.registerBreakpoints();
+            _.init(true);
             _.checkResponsive(true);
 
         }
@@ -572,7 +551,7 @@
 
     };
 
-    Slick.prototype.checkResponsive = function(initial) {
+    Slick.prototype.checkResponsive = function(initial, forceUpdate) {
 
         var _ = this,
             breakpoint, targetBreakpoint, respondToWidth, triggerBreakpoint = false;
@@ -587,8 +566,9 @@
             respondToWidth = Math.min(windowWidth, sliderWidth);
         }
 
-        if (_.originalSettings.responsive && _.originalSettings
-            .responsive.length > -1 && _.originalSettings.responsive !== null) {
+        if ( _.options.responsive && 
+            _.options.responsive.length && 
+            _.options.responsive !== null) {
 
             targetBreakpoint = null;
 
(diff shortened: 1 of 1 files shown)


slick/slick.js
```diff
@@ -1472,6 +1454,46 @@
 
     };
 
+    Slick.prototype.registerBreakpoints = function() {
```
kenwheeler: Can you annotate this? Is it thoroughly tested?

slick/slick.js
```diff
@@ -1472,6 +1454,46 @@
...
+
+        var _ = this, breakpoint, currentBreakpoint, l,
+            responsiveSettings = _.options.responsive || null;
+
+        if ( $.type(responsiveSettings) === "array" && responsiveSettings.length ) {
```
simeydotme: just a sanity check to make sure `_.options.responsive` is actually a usable array and not a retard value.

slick/slick.js
```diff
@@ -1472,6 +1454,46 @@
...
+            for ( breakpoint in responsiveSettings ) {
+
+                l = _.breakpoints.length-1;
+                currentBreakpoint = responsiveSettings[breakpoint].breakpoint;
+
+                if (responsiveSettings.hasOwnProperty(breakpoint)) {
+
+                    // loop through the breakpoints and cut out any existing
+                    // ones with the same breakpoint number, we don't want dupes.
+                    while( l >= 0 ) {
```
simeydotme: Check each breakpoint inside the `_.options.responsive` array,
Check that the current breakpoint we're registering (lets say `800`) doesn't already exist anywhere in teh `_.options.responsive` option. -- if it does, splice it out of the array.  

---
Basically, I tried it without this, and noticed that if I supplied an array with a breakpoint which already exists, then it would just take the first one it found in the array when we get to the `checkResponsive()` method. -- so the decision is that if I'm adding a breakpoint of `800` then, I do not want any current breakpoints which are also at `800` as only the first one would trigger, anyway. .. and I just added a new one, so ofcourse; I want that one? #logic hehe.

slick/slick.js
```diff
@@ -1489,22 +1511,19 @@
             _.currentSlide = 0;
         }
 
-        _.setProps();
+        _.registerBreakpoints();
```
simeydotme: this allows us to re-register the breakpoints when the `setOption()` method runs.

slick/slick.js
```diff
@@ -1680,8 +1699,27 @@
...
     Slick.prototype.setOption = Slick.prototype.slickSetOption = function(option, value, refresh) {
 
-        var _ = this;
-        _.options[option] = value;
+        var _ = this, l, item;
+
+        if( option === "responsive" && $.type(value) === "array" ) {
+            for ( item in value ) {
+                if( $.type( _.options.responsive ) !== "array" ) {
+                    _.options.responsive = [ value[item] ];
```
simeydotme: if originally the `_.options.responsive` was `null` then we want to wrap the current `object` in an array. -- otherwise we want to remove any duplicate breakpoints, and then push the current breakpoint.

slick/slick.js
```diff
@@ -587,8 +566,9 @@
             respondToWidth = Math.min(windowWidth, sliderWidth);
         }
 
-        if (_.originalSettings.responsive && _.originalSettings
-            .responsive.length > -1 && _.originalSettings.responsive !== null) {
+        if ( _.options.responsive && 
+            _.options.responsive.length && 
+            _.options.responsive !== null) {
```
simeydotme: this was wroing, using `_.originalSettings.responsive` as now the options can be set post-initilisation so we need to check if responsive was added to `_.options`

slick/slick.js
```diff
@@ -144,27 +143,6 @@
...
-                    }
-                }
-                _.breakpoints.sort(function(a, b) {
-                    if (_.options.mobileFirst === true) {
-                        return a - b;
-                    } else {
-                        return b - a;
-                    }
-                });
-            }
```
simeydotme: remove the breakpoint registration from setup, and place in to it's own method with de-duplication

slick/slick.js
```diff
@@ -572,7 +551,7 @@
 
     };
 
-    Slick.prototype.checkResponsive = function(initial) {
+    Slick.prototype.checkResponsive = function(initial, forceUpdate) {
```
simeydotme: without `forceUpdate` the `checkResponsive` method would never be able to refresh the slider when someone adds breakpoints via `setOption()` method.

simeydotme:
Added some annotations, for you, @kenwheeler :)
https://github.com/kenwheeler/slick/issues/327#issuecomment-100597175
^^ this comment has a pretty thorough JSFiddle attached, and I tested-the-mother-loving-shit out of it on my local machine. -- the commits belie the simplicity of the feature, really... the large chunks are just de-duping the `_.options.responsive` array and pushing new values :) -- the rest of it is just code-rearrangement and it was tested well. :sweat_smile:

kenwheeler:
Thank you so much for the annotations. I need to fiddle with this myself a bit before merging, cause this is a biggum. Well done.

simeydotme:
Yeh absolutely mate! - cheers, and it would have been much simpler to de-dupe if the responsive JSON-style array was an indexed-object instead, but I had to work around that architecture 😙 ... Perhaps something to consider for v2 😄

kenwheeler:
:+1:

PR #1344, opened 2015-05-13
changing zIndex magic-number in fadeSlilde() and setFade() to a user-declared option variable

nominalaeon:
PROMPT: our site has a drop down menu whose z-index was being trumped
when we had Slick fade set to true. Rather than adjust and have to
regression test our component, I wanted Slick to be flexible enough to
accommodate pre-existing components’ z-indexes when it’s being
integrated later in a project.

CHANGES: zIndex for currentSlide was being set to “1000”. I added
zIndex to the set of author-able options and set it to 1000 by default.

The extra level of control is non-invasive as the current zIndex
settings are relative to one another. Since none of the other Slick
elements have z-indexing authors couldn’t break Slick itself.

I changed the differences to be single instead of by hundreds since a
user could potentially submit a single-digit zIndex.

Since the setFade() zIndex has to drop by 2, there is a check at the
beginning to make sure the zIndex can never go to 0 or -1 by setting it
to at least 3.

PAIN-POINT: I didn’t put in a check to see if z-index was defined as a
valid number, but I’m not seeing where any other variable is validated
so that seemed okay.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -85,7 +85,8 @@
                 variableWidth: false,
                 vertical: false,
                 verticalSwiping: false,
-                waitForAnimate: true
+                waitForAnimate: true,
+                zIndex: 1000
             };
 
             _.initials = {
@@ -164,6 +165,10 @@
                         return b - a;
                     }
                 });
+            }
+
+            if (_.options.zIndex < 3) {
+                _.options.zIndex = 3;
             }
 
             if (typeof document.mozHidden !== 'undefined') {
@@ -885,7 +890,7 @@
         if (_.cssTransitions === false) {
 
             _.$slides.eq(slideIndex).css({
-                zIndex: 1000
+                zIndex: _.options.zIndex
             });
 
             _.$slides.eq(slideIndex).animate({
@@ -898,7 +903,7 @@
 
             _.$slides.eq(slideIndex).css({
                 opacity: 1,
-                zIndex: 1000
+                zIndex: _.options.zIndex
             });
 
             if (callback) {
@@ -1648,7 +1653,7 @@
                     position: 'relative',
                     right: targetLeft,
                     top: 0,
-                    zIndex: 800,
+                    zIndex: _.options.zIndex - 2,
                     opacity: 0
                 });
             } else {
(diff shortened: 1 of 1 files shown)


simeydotme:
Nice idea and explanation, @nominalaeon ! :)
If i could ask you kindly to provide a working fiddle to demonstrate it (using your unminified source) I will review and pull! 

Ace work.
Simon.

nominalaeon:
Thanks, Simon!

Chrome and FireFox are smart enough to handle the z-index layering. This is totally an IE thing, but I think I can make a case for z-index control in general besides.

I'm working up a Codepen and I'll get it up soon.

Noah

simeydotme:
I agree in general about z-index control.
So many websites have select-box polyfills, and modal-dialogs now... and if there's any discrepancies with positioning, or layering it can really cause a headache when a 3rd-party module has set crazy z-indexing. 

Your code reads cleanly and is easy to follow and dare I say it compiles in my head pretty perfectly, but I'd like to throw it some scenarios to deal with before accepting it ;D

Si.

nominalaeon:
Fire up your IE8 VM!

http://s.codepen.io/nominalaeon/debug/doMaRL?

This is a simplified example of what we encountered. The top example is using the default zIndex setting and the bottom example sets zIndex to 3 with data-slick='{"zIndex": 3}'. Obviously this could be set in the JavaScript Slick options just as easy.

Our problem was our Nicolas Cage is set and tested at z-index 600, so Slick's default 1000 was unintentionally covering it up.

Modern browsers are too smart for this fix to have a use case outside of IE8, so making the z-index author-able might be a slightly overkill solution. I'm guessing that since the Slick container itself isn't getting z-indexed, modern browsers know not to let its children screw with the layer order of other page elements.

Let me know if more/different examples are required, and thanks for all your help, Simon.

simeydotme:
Excellent!
Actually, I will have to double confirm but i believe it's because the W3 spec states z-indexes are "relative" to their container if they are positioned. 

But IE don't be givin' no damn about pesky specs!

I'll research a bit, I also saw some old issues and requests where others tried to change the zindexing and @kenwheeler agreed, but must have gotten too busy to implement.

I personally like your solution, as the website I maintain had many models and dropdowns using `z-index`, so i wont have to revise them all.

simeydotme:
> The root element forms the root stacking context. Other stacking contexts are generated by any positioned element (including relatively positioned elements) having a computed value of 'z-index' other than 'auto'. Stacking contexts are not necessarily related to containing blocks. In future levels of CSS, other properties may introduce stacking contexts, for example 'opacity' [CSS3COLOR].

http://www.w3.org/TR/CSS21/visuren.html#z-index

Issue #1342, opened 2015-05-13
Fade option buggy when content is different sizes


danieldunderfelt:
Hi!

I came across this when trying to use slick on a bunch of images with different sizes without their own slide background:

http://jsfiddle.net/s43ybj7x/4/

When the fade option is enabled and some slides' content is smaller than the previous, the previous content stays in the background and disappears after a split second without a transition. Basically only the new content has the fade effect, while the old content disappears once the fade is complete.

Would it be possible to enable the fade effect for the outgoing slide as well?

Thanks!

ethanclevenger91:
Was running into this issue today as well with items that don't have a background image and are rather just text.

http://codepen.io/ethanclevenger91/pen/bdpmLg

Created a pull request

ethanclevenger91:
This wasn't in the latest hotfix (or at least not in the minified version) - any way we can get a new build? I don't know what the procedure is for this so figured I'd just petition.

walpek:
+1 Would love to see this issue fixed

kenwheeler:
Getting a release pushed soon, sorry guys

PR #1349, opened 2015-05-14
Fade out previous slide

ethanclevenger91:
Here's the bug in question for #1342 : http://jsfiddle.net/7pvu8gv2/
Fixed: http://jsfiddle.net/m4dL92nq/

Another manifestation: http://codepen.io/ethanclevenger91/pen/PqNxOO
Another fixed: http://codepen.io/ethanclevenger91/pen/jPqQxQ

To say this was rigorously tested might be an overstatement. Let me know if updates should be made.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -914,6 +914,37 @@
 
     };
 
+    Slick.prototype.fadeSlideOut = function(slideIndex, callback) {
+
+        var _ = this;
+
+        if (_.cssTransitions === false) {
+
+            _.$slides.eq(slideIndex).animate({
+                opacity: 0
+            }, _.options.speed, _.options.easing, callback);
+
+        } else {
+
+            _.applyTransition(slideIndex);
+
+            _.$slides.eq(slideIndex).css({
+                opacity: 0
+            });
+
+            if (callback) {
+                setTimeout(function() {
+
+                    _.disableTransition(slideIndex);
+
+                    callback.call();
+                }, _.options.speed);
+            }
+
+        }
+
+    };
+
     Slick.prototype.filterSlides = Slick.prototype.slickFilter = function(filter) {
 
         var _ = this;
@@ -1993,6 +2024,9 @@
 
         if (_.options.fade === true) {
             if (dontAnimate !== true) {
+                _.fadeSlideOut(oldSlide, function() {
+                    _.postSlide(oldSlide);
+                })
                 _.fadeSlide(animSlide, function() {
                     _.postSlide(animSlide);
                 });
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -884,10 +884,6 @@
 
         if (_.cssTransitions === false) {
 
-            _.$slides.eq(slideIndex).css({
-                zIndex: 1000
-            });
-
             _.$slides.eq(slideIndex).animate({
                 opacity: 1
             }, _.options.speed, _.options.easing, callback);
@@ -897,8 +893,7 @@
             _.applyTransition(slideIndex);
 
             _.$slides.eq(slideIndex).css({
-                opacity: 1,
-                zIndex: 1000
+                opacity: 1
             });
 
             if (callback) {
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -884,6 +884,10 @@
 
         if (_.cssTransitions === false) {
 
+            _.$slides.eq(slideIndex).css({
+                zIndex: 1000
+            });
+
             _.$slides.eq(slideIndex).animate({
                 opacity: 1
             }, _.options.speed, _.options.easing, callback);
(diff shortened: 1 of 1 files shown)


simeydotme:
@ethanclevenger91 ... could you take a quick look over: https://github.com/kenwheeler/slick/pull/1344 ( https://github.com/nominalaeon/slick/commit/73c30873295b450f32a417f318b20940bee7da59 ).. does your PR mean that these z-index changes are unecessary, also you can see from @nominalaeon s commits that there's more z-index references than the ones you have erased, could you make sure those other z-index references are not conflicting, or affecting your PR please.
:smile: 

Thanks , Simon.

nominalaeon:
the setFade() function (line 1665) is what makes fade transitions work if you're using the setPosition method to jump between slides. There the z-index ensures that the incoming slide is definitely behind the outgoing no matter what their mark-up order.

It does look like z-index is benign in both setFade() and fadeSlide() but, like I mentioned in #1344, I only noticed the z-index stuff in IE. The z-index declarations are probably just harmless safe-guards for IE compatibility. Chrome and FireFox are deceivingly competent in this department.

But someone's going to have to test in IE8 or 9 to confirm and I'm calling not-it right now.

ethanclevenger91:
My original commit did the trick and surely the z-indexs were there initially for some reason, IE being a likely culprit. No reason I couldn't/shouldn't just revert. Then you guys can continue to address #1344 however you'd planned. Does that make sense?

If we want to be sure, I should have a VM handy for testing IE8.

nominalaeon:
The fix for #1344 is to make zIndex an arg you can pass in. There might be potential use cases for controlling the slide z-index, but that kind of functionality is totally scope creep for a slider.

If the transitions work fine without z-index being set for IE8 (or just IE9? Ken or Simon would have to speak to that), it would be worth dropping them out from both functions.

simeydotme:
Hi @ethanclevenger91 , actually that's not what I was expecting :smile: -- your code addresses a rather prevalent issue which keeps getting raised, so I was hoping to merge it in! :smile: 

But I wanted to make sure we cover those extra few `z-index` assignments:
- https://github.com/ethanclevenger91/slick/blob/629cd590bdd88cbf58589a0da6b24a851e3d9044/slick/slick.js#L1677 
- https://github.com/ethanclevenger91/slick/blob/629cd590bdd88cbf58589a0da6b24a851e3d9044/slick/slick.js#L1685
- https://github.com/ethanclevenger91/slick/blob/629cd590bdd88cbf58589a0da6b24a851e3d9044/slick/slick.js#L1692

If the z-indexes are un-needed now, we can remove them... 
@nominalaeon PR is to allow control of the z-indexes... so if your PR removes the z-indexes, then there's no need for #1344. :) ..... 

__Thus__ my question is: Could you investigate removing those other `zIndex` in your branch? :bow: 

Si.

nominalaeon:
I will, actually I'll just take them out and see if the issue we had on our site is still resolved. brb

simeydotme:
Cool, thanks so much! -- and don't forget to test all the scenarios where `setPosition()` is called :disappointed_relieved:

nominalaeon:
:\ Well without the z-index the slides do not cover my dropdown menu, however, they don't operate at all in IE8. Which is just a terrible thing to say out loud. on the internet. but the client wants compatibility. 

Someone should double-check these results, but what I'm seeing is the last slide on top at full opacity. The dot navigation is blinking between slides, but the last slide is the only one that's ever visible. 

Works fine in IE9 though! I have a slider that runs on a nav-tabs aside that uses setPosition() to jump around and I have the exact same results with it. Borked in IE8, amazing everywhere else.

I have my hack in place in my client's code-base, so that's covered. Ethan's change (when it updates to remove setFade() z-index'ing too) gets my vote. If you want IE8 compatibility, be warned fade-transition is not available. Seems fair to me.

ethanclevenger91:
So sounds like if I clear the rest of the z-index's we...
**Gain:**
Less intrusive on modern browsers
**Lose:**
IE8 compatibility for fade transitions.

I'm fine with that, but want to make sure before I update the commit.

nominalaeon:
ugh, there are a lot of IE8 mentions in the Issues though.

Ethan, does having the z-index property author-able help your situation? We could make a "none" setting too to completely ignore z-index.

ethanclevenger91:
It doesn't affect my _original_ issue at all - like I said, I can revert to the original commit I made that doesn't alter the z-index'ing. If we wanted to add a passable z-index option, I would propose addressing that in another update.

simeydotme:
Alright, no way we can f:cyclone:k up IE8 support, so if you can revert your `zIndex` commit, I'll give this PR a proper testing with a bunch of options and scenarios and hopefully merge it soon... then I'll work with @nominalaeon to rebase and test his branch, too :smiley: 

Thanks for the effort, guys! :bow:
Si.

PR #1357, opened 2015-05-17
Fixing #1353 by filtering the slides in buildOut()

### LINKED ISSUE #1353: Not limiting Slick children to visible tags.
https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L491-L493 is checking against slick-cloned, but it's ignoring the options.slide setting.

Fixed in https://jsfiddle.net/simeydotme/50kp465g/2/ but not regression tested.

simeydotme:
This filter was removed at
https://github.com/kenwheeler/slick/commit/9d13d6abc4ab4e05948b5bab47b2574fad9c9c64#diff-eaaa72f2e5f1d4974e4f1ae902717908L476
for some reason, I cannot figure out why. Since it was removed, filtering
slides was not working on initialization.

Tested against Init.
Tested against setOption.
Tested against breakpoints.

Fixes #1353

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -489,9 +489,11 @@
 
         var _ = this;
 
-        _.$slides = _.$slider.children(
-            ':not(.slick-cloned)').addClass(
-            'slick-slide');
+        _.$slides = 
+            _.$slider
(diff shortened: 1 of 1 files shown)


kenwheeler:
:+1:

PR #1361, opened 2015-05-18
Prevent double-binding the click event

cohenudi:
Prevent double-binding the click event to the next/prev arrows. This scenario happens when you use the removeSlide method.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1180,14 +1180,16 @@
         var _ = this;
 
         if (_.options.arrows === true && _.slideCount > _.options.slidesToShow) {
-            _.$prevArrow.off('click.slick');
-            _.$prevArrow.on('click.slick', {
-                message: 'previous'
-            }, _.changeSlide);
-            _.$nextArrow.off('click.slick');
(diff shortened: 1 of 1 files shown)


simeydotme:
Thanks, I'll check this out, actually it's happening in the `reinit()` method, I'd imagine this should be happening in quite a few instances! :confounded:

cohenudi:
I can imagine there are a lot more scenarios than the one I mentioned exactly because of what you said (the `reinit()`). Good luck with the investigation :grin:

simeydotme:
yo, @cohenudi ... sorry for the inactivity here. :disappointed: 

I was hoping you could rearrange the code slightly to be:
`.off(...).on(...)` 

And if at all possible use the fiddle http://jsfiddle.net/simeydotme/fmo50w7n/ as a base to show it working before/after? :)

I appreciate if you're busy, but I'd love to merge your code asap! :grin: :

cohenudi:
Hey :-) 
I made the changes and chained the methods. I actually like method chaining much better and gonna use it when possible from now. Thanks for that.

PR #1376, opened 2015-05-21
wrong extend order in refresh function

janrode:
refresh function: wrong extend order for current settings _ and _.initials.

this will break settings like initialSlide in combination with responsive.
in this case the initialSlide property will be overwritten by _.initials.initialSlide which is always 0.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1489,7 +1489,7 @@
 
         _.destroy(true);
 
-        $.extend(_.initials, _);
+        $.extend(_, _.initials, { currentSlide: currentSlide });
 
         _.init();
 
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1489,7 +1489,7 @@
 
         _.destroy(true);
 
-        $.extend(_, _.initials);
+        $.extend(_.initials, _);
 
         _.init();
 


simeydotme:
Jan, can you produce a JSFiddle to demonstrate the error using the example JSfiddle provided in the contribution guidelines... Also can you demonstrate it being fixed after with your patch... And can you provide some edge cases, like `asNavFor` and `addSlide` and `setOption`.

Thanks.

janrode:
Hi,
- ERROR Example: http://jsfiddle.net/rf8v98n7/1/ 
after resizing the window run the fiddle again and if the width is > 639, the initalSlide is ignored.

- FIXED Example: http://jsfiddle.net/f8vonv89/1/

simeydotme:
Thanks a ton for creating those fiddles!! :)
You have found a bug, but your patch is incorrect :cry: -- actually what is happening is the `currentSlide` is being reset wrongly... the `initialSlide` is maintaining it's integrity correctly, but it is not being used upon `breakpoint`, as when we hit a `breakpoint` we want to use the `currentSlide`, not the `initialSlide`. :)

please change your code to:

```js
$.extend(_, _.initials, { currentSlide: currentSlide });
```

and I will accept the PR :)

janrode:
Thank you for your assistance! 
Bugfix is committed! Hopefully correct, this time :)

simeydotme:
Ya, no need to worry, Slick is a big thing, so gotta be really focussed on what you may affect with changing code :) -- theres a hundred options to consider and how they interact with each property! :smile:

Issue #1421, opened 2015-06-05
Dynamically added slides break internal index counter


MaxKorz:
##### Steps to Reproduce:
1. Go to JSBin http://jsbin.com/mademegize/3/edit?html,js,output
2. Click Next

##### Expected Results: 
* .slick-active stays .slick-active

##### Actual Results:
* Green (.slick-active) turns red (.slick)

##### What caused this:
* New slide was added when currentSlide != 0

kenwheeler:
Noted, closing pending PR merge

PR #1422, opened 2015-06-05
Make sure slick-active is currentSlide after reinit()

MaxKorz:
Fixes https://github.com/kenwheeler/slick/issues/1421

>##### Steps to Reproduce:
>1. Go to JSBin http://jsbin.com/mademegize/3/edit?html,js,output
>2. Click Next
>
>##### Expected Results: 
>* .slick-active stays .slick-active
>
>##### Actual Results:
>* Green (.slick-active) turns red (.slick)
>
>##### What caused this:
>* New slide was added when currentSlide != 0

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1545,7 +1545,7 @@
             $(_.$slideTrack).children().on('click.slick', _.selectHandler);
         }
 
-        _.setSlideClasses(0);
+        _.setSlideClasses(typeof _.currentSlide === 'number' ? _.currentSlide : 0);
 
         _.setPosition();
 


simeydotme:
@MaxKorz -- this looks good at first glance, but your JSBin is lacking a lot of test cases... could you please create a JSFiddle that looks something more like: http://jsfiddle.net/0u5qyg8x/ -- there's hundreds of options and edge-cases, so I need to see at least some evidence it works as expected across many different configurations. 

If you could duplicate the fiddle as well (one before fix, one after) that'd be great.

Regards, Simon.

MaxKorz:
Sure! Will do it tomorrow

MaxKorz:
Before: http://jsfiddle.net/f6b6qt14/1/
After: http://jsfiddle.net/o4rhaff3/1/

Btw is it okey that all carousels (even those that have autoplay: false) slide automatically?

And I can see that Third carousel doesn't have slick-active class

simeydotme:
they are all auto-playing because of "asNavFor" :)
I'll take a closer look another time at your code.

Si

microadam:
Any chance of this being merged anytime soon? Hit the same issue

kenwheeler:
@microadam can you confirm this works?

microadam:
It works in the situation I am using it in. Can't confirm that it doesn't introduce other issues though

kenwheeler:
Ok I'll run it through the paces

PR #1438, opened 2015-06-10
Pass dontAnimate through to asNavFor

MrAvenger:
Is it possible for dontAnimate to be passed into the NAV as I don't wont either of them to animate (for this one time only).
I've tried doing it like this:
```
$('.slick').slick('slickGoTo', 1, true);
$('.slick-nav').slick('slickGoTo', 1, true);
```
Because the animation is already under way from the the first call it doesn't take effect.

Thanks.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -338,7 +338,7 @@
 
     };
 
-    Slick.prototype.asNavFor = function(index) {
+    Slick.prototype.asNavFor = function(index, dontAnimate) {
 
         var _ = this,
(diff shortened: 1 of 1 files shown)


simeydotme:
@MrAvenger your request is sensible.
Apologies for the lack of response until now. The repo went through a period of low activity.

Please could you rebase your code with master, and provide a [jsfiddle like this one (fork it)](http://jsfiddle.net/simeydotme/fmo50w7n/) to demonstrate the working change?

I'd :heart: to merge.
​

PR #1146, opened 2015-03-16
Accessibility

msrafi:
http://jsfiddle.net/1x8w9c3n/31/

PR #838, opened 2014-11-28
Screen Reader and Keyboard Accesibility

msrafi:
Added Following Functionality

1. Screen Reader Attributes
2. Keyboard Accessibility
3. Focus Handling in Auto play mode

Issue #792, opened 2014-11-16
Positioning issue when tabbing through items with links


robinpoort:
When slider items contain links and when you tab through them the positioning of items gets all screwed up. Depending on browsers the position gets recovered or not screwed at all.

JSFiddle: http://jsfiddle.net/tu2xqtej/4/

Safari 8: Nothing gets screwed because the tabindex never reaches links within slider items. Not very accessible
Chrome 38: position gets screwed and does not recover when sliding again (with arrow or arrow keys)
Firefox 33: position get screwed but recovers when sliding again (with arrow or arrow keys)
IE11, IE10 and IE9: Behave the exact same as Chrome does

I'm sure the tabindex behaviour is not something you can really do anything about since it's browser behaviour.

I think some other slider libraries don't have this problem because they are positioning items with CSS "position" either relative or absolute where as Slick is using floats to position items next to each other.

How to solve this?
Using CSS "position" might come to the rescue. However, if you tab through the items the link will be selected but you will see nothing and it would appear if the 'tab' button isn't doing anything. At that point it's cycling through all the "invisible" positioned items. Not very accessible.

I think a better approach would be to see if you can listen (with JS) to the tabindex and have the slider animate when the focus is on a tab in a slide. However this would also mean that if you have three sliding items and there's only a link in the first and last item it would skip the rest and go to the last item immediately. This is also what happens in the second slider in my JSFiddle example. To fix that you would have to add a "tabindex" to every slide so you could easily always slide through each item. That's what I did in the third slider. However you would always have to go through each slide before you could continue when just using the keyboard. and this method is also not working in Safari.

Not sure if solvable and what the best solution would be. But thought I'd at least let you know that I ran into this problem and some ways I thought of solving this.

msrafi:
Guys, I spent some time since last weekend and found the changes that I am making is affecting the performance when using it with breakpoints/resize. I would consider this would be a good to have feature with design dependency. But, we can close the accessibility issue and open it as a new feature. Will try later when I get more time.

robinpoort:
Agreed let's handle this in a different issue:
https://github.com/kenwheeler/slick/issues/912

MattDiMu:
@kenwheeler  Any chance the tabindex change is merged soon, so using tab-index doesn't destroy the slider anymore? Willing to help if needed.

Cottser:
This seems to be the relevant PR: #1146

camdagr8:
[...]

PR #1467, opened 2015-06-22
Accessibility Updated

msrafi:
JS Fiddle added: http://jsfiddle.net/equinoxs/1x8w9c3n/39/

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2552,14 +2552,3 @@
     };
 
 }));
-
-
-$(function(){
-   $('.responsive').slick({
-            dots: true,
-            infinite: true,
-            speed: 500,
-            slidesToShow: 1,
-            slidesToScroll: 1            
-        }); 
-});
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -45,7 +45,7 @@
                 appendDots: $(element),
                 arrows: true,
                 asNavFor: null,
-               prevArrow: '<button type="button" data-role="none" class="slick-prev" aria-label="Previous" tabindex="0" role="button">Previous</button>',
+                prevArrow: '<button type="button" data-role="none" class="slick-prev" aria-label="Previous" tabindex="0" role="button">Previous</button>',
                 nextArrow: '<button type="button" data-role="none" class="slick-next" aria-label="Next" tabindex="0" role="button">Next</button>',
                 autoplay: false,
                 autoplaySpeed: 3000,
@@ -526,10 +526,6 @@
 
         _.updateDots();
 
-        // This is controlled through _.activeADA method
-        //if (_.options.accessibility === true) {
-        //    _.$list.prop('tabIndex', 0);
-        //}
 
         _.setSlideClasses(typeof _.currentSlide === 'number' ? _.currentSlide : 0);
 
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -45,15 +45,15 @@
                 appendDots: $(element),
                 arrows: true,
                 asNavFor: null,
-                prevArrow: '<button type="button" data-role="none" class="slick-prev" aria-label="previous">Previous</button>',
-                nextArrow: '<button type="button" data-role="none" class="slick-next" aria-label="next">Next</button>',
+               prevArrow: '<button type="button" data-role="none" class="slick-prev" aria-label="Previous" tabindex="0" role="button">Previous</button>',
+                nextArrow: '<button type="button" data-role="none" class="slick-next" aria-label="Next" tabindex="0" role="button">Next</button>',
                 autoplay: false,
                 autoplaySpeed: 3000,
                 centerMode: false,
                 centerPadding: '50px',
                 cssEase: 'ease',
                 customPaging: function(slider, i) {
-                    return '<button type="button" data-role="none">' + (i + 1) + '</button>';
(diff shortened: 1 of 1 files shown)


msrafi:
@kenwheeler 
Hi, please let me know if this works. Comments are welcome.

slick/slick.js
```diff
@@ -526,9 +526,10 @@
 
         _.updateDots();
 
-        if (_.options.accessibility === true) {
-            _.$list.prop('tabIndex', 0);
-        }
+        // This is controlled through _.activeADA method
+        //if (_.options.accessibility === true) {
+        //    _.$list.prop('tabIndex', 0);
+        //}
```
oliverbenns: If it's not being used, let's get rid of it, not comment it out.

slick/slick.js
```diff
@@ -45,15 +45,15 @@
                 appendDots: $(element),
                 arrows: true,
                 asNavFor: null,
-                prevArrow: '<button type="button" data-role="none" class="slick-prev" aria-label="previous">Previous</button>',
-                nextArrow: '<button type="button" data-role="none" class="slick-next" aria-label="next">Next</button>',
+               prevArrow: '<button type="button" data-role="none" class="slick-prev" aria-label="Previous" tabindex="0" role="button">Previous</button>',
```
oliverbenns: Slight indentation fix needed here :+1:

oliverbenns:
Nice work dude!

msrafi:
comments removed

msrafi:
Thanks @oliverbenns

kenwheeler:
For some reason I just saw this,I'll review tonight, thanks!

msrafi:
History: #838 #1146

slick/slick.js
```diff
@@ -2473,3 +2552,14 @@
     };
 
 }));
+
+
+$(function(){
```
kenwheeler: What's this all about?

slick/slick.js
```diff
@@ -1283,19 +1284,21 @@
...
-                }
-            });
-        } else if (event.keyCode === 39 && _.options.accessibility === true) {
-            _.changeSlide({
-                data: {
-                    message: 'next'
-                }
-            });
+         //Dont slide if the cursor is inside the form fields and arrow keys are pressed
+        if(!event.target.tagName.match('TEXTAREA|INPUT|SELECT')) {
```
jandersonBB: ++ nice

slick/slick.js
msrafi: My Bad. Kept the sample testing initiation. Removed it now.

slick/slick.js
kenwheeler: nice indeed

kenwheeler:
Well done @msrafi

msrafi:
Cool. Thank you

Cottser:
:clap: thanks everyone! So happy to see this in.

leggomuhgreggo:
Hey, the addition of the `tabindex` attribute is triggering the focus state when a slide has `tabindex="0"`. This is causing the default outline from the user agent styles around the slide.

It turns out that its happening on Chrome 43 [[screenshot](http://i.imgur.com/6VpGsWp.png)], and IE11, but not Firefox 39. I think Firefox may have depricated support at some point, becuase in looking into the issue it seems like they [had it at one point](https://nemisj.com/focusable/), but [don't now](http://stackoverflow.com/questions/30883332/tabbing-throgh-keyboard-not-working-in-mozilla-firefox).

I'm not sure what the best practice is in situations like this. Clearly the focus state is useful for accessibility, but it's obtrusive from a design standpoint. 

I think I'm probably going to add  `.slick-slide:focus{outline:0}` in production, and I wonder if adding it to slick-theme.css a fair middle ground?

[Here's an example](http://codepen.io/gwst/pen/Xbqymp) with the style override commented out.

Thanks!

msrafi:
I added this for screen reader. So when it slides the screen reader will actually read the details about that particular slide. Otherwise there is no option for letting the screen reader to give that input. 

I agree on the chrome outline, but I would recommend leaving the slick css very generic so that the CSS can be updated specific to project requirements.

You can read more #792 on this topic.

Thanks

leggomuhgreggo:
@msrafi So having `outline: 0` in the slick-theme.css but not in slick.css makes sense to you too, then?

After skimming #792 I can see there was definitely some thought put into the tab focus sequencing, but I noticed that when there are multiple slides, the focus state is added to the right-most of the slick-active slides. Might be I'm missing something, but does it made more sense to add the focus state to the first (left-most) of the slide-active slides--at least when the RTL option is false?

I was messing around, and I'm pretty sure just adding `.eq(0)` to the slick-active selection on line [2599](https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L2599) does the trick, where, after scroll, the left-most slide is first to be focused, then subsequent visible ones then next nav. 

Might be worth conditioning based on RTL. That functionality seems messy at the moment, though.

Based on reading the spec, I actually tried to order the tabindexes sequentially so that the left-most slide would have `tabindex=0` and the next slide `tabindex=1` and so on, but, at least in Chrome, the tab sequence went from `tabindex=0` right to the next arrow instead of `tabindex=1`. Not sure why it behaves that way, but I guess all of them having `tabindex=0` works fine.

Thanks

Issue #1466, opened 2015-06-22
Incorrect amount of dots


danielpuglisi:
When using `centerMode: true` and anything higher than `slidesToShow: 1` the amount of dots displayed is wrong.

Here is a fiddle: http://jsfiddle.net/4u4rvehe/5/

Any ideas on how to fix this?

timonf:
Same or similar issue appears on slick demo page (https://kenwheeler.github.io/slick/):

*  See "Autoplay" when 5, 6, 1 or 6, 1, 2 are visible.
*  See "Slider Syncing" when 3, 4, 5 or 4, 5, 1 are visible.

A "workaround" could be to set number of slidesToShow to the same as slidesToScroll.

leggomuhgreggo:
Been fixed in the master. [Here's a fork with the updated js](http://jsfiddle.net/leggomuhgreggo/ztd7zhLa/). Thanks!

PR #1471, opened 2015-06-23
Fix wrong amount of dots in centerMode

danielpuglisi:
Fixes issue #1466

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -975,14 +975,14 @@
         var counter = 0;
         var pagerQty = 0;
 
-        if (_.options.infinite === true) {
+        if (_.options.centerMode === true) {
+            pagerQty = _.slideCount;
+        } else if (_.options.infinite === true) {
             while (breakPoint < _.slideCount) {
                 ++pagerQty;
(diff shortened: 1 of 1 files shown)


apetrakow:
does this fix work with `centerMode: false`? 
I have a pretty similar issue here: http://jsfiddle.net/0bnor44n/

danielpuglisi:
No.

You need to set `infinite: false`: http://jsfiddle.net/0bnor44n/1/

If you want it to work with infinite I think you need to change the following lines: https://github.com/danielpuglisi/slick/blob/b7ef77ffb0ef26734d0362df614793bce5ec94e0/slick/slick.js#L981-L985

nikitaaswani:
@danielpuglisi I am also facing the similar issue. I am using centerMode with `infinite: false` and it shows me number of slides + 1 dots. Were you facing the same issue?

danielpuglisi:
@nikitaaswani I don't remember, I don't use slick anymore. It doesn't seem to be maintained anymore anyway.

I'm currently using https://github.com/nolimits4web/swiper in case you need a good alternative slider ;)

PR #1492, opened 2015-06-29
When in accessibility mode, carousel should not change panel when edi…

jandersonBB:
…tting an input.  Need to preserve arrow functionality for navigating the text field

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1284,6 +1284,8 @@
 
         var _ = this;
 
+        if ($("input:focus").length !== 0) return;
+
         if (event.keyCode === 37 && _.options.accessibility === true) {
             _.changeSlide({
                 data: {


ermterm:
From WCAG:

3.2.1 On Focus: When any component receives focus, it does not initiate a change of context. (Level A)
How to Meet 3.2.1 | Understanding 3.2.1
3.2.2 On Input: Changing the setting of any user interface component does not automatically cause a change of context unless the user has been advised of the behavior before using the component. (Level A)

jandersonBB:
From WCAG:

3.2.1 On Focus: When any component receives focus, it does not initiate a change of context. (Level A)
How to Meet 3.2.1 | Understanding 3.2.1
3.2.2 On Input: Changing the setting of any user interface component does not automatically cause a change of context unless the user has been advised of the behavior before using the component. (Level A)

msrafi:
#1467 Covered this accessibility update

PR #1509, opened 2015-07-06
update changeSlide to preventDefault if target is a child of an a tag

### LINKED ISSUE #1508: Next / Prev links don't apply preventDefault if they have sub-elements
Using the following markup for the next/prev buttons causes the page to jump when the next/prev buttons are clicked because it skips over the `event.preventDefault();` code since the target element is not an `a` tag.

    nextArrow: '<a href="#" class="slick-next"><i class="sprite rightArrowIcon"></i></a>',
    prevArrow: '<a href="#" class="slick-prev"><i class="sprite leftArrowIcon"></i></a>',

The issue can be seen here at the following JSFiddle (note: due to the nature of the issue, I have to link to the iframe code directly, rather than to the fiddle itself. Note that the URL gets a `#` in it when clicking next/prev.) 

Bugged JSFiddle:
https://fiddle.jshell.net/localpcguy/vbyg5wxx/1/show/light/
original: http://jsfiddle.net/localpcguy/vbyg5wxx/1/

Will shortly have an incoming fix that was used to solve this by using `event.currentTarget` rather than `event.target`. 

Fixed JSFiddle:
https://fiddle.jshell.net/localpcguy/vbyg5wxx/2/show/light/
original: http://jsfiddle.net/localpcguy/vbyg5wxx/2/

localpcguy:
update changeSlide to preventDefault even if target is a child of an `<a>` tag

bugged jsfiddle:
- https://fiddle.jshell.net/localpcguy/vbyg5wxx/1/show/light/
- http://jsfiddle.net/localpcguy/vbyg5wxx/1/

working jsfiddle:
- https://fiddle.jshell.net/localpcguy/vbyg5wxx/2/show/light/
- http://jsfiddle.net/localpcguy/vbyg5wxx/2/

Fixes #1508

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -668,17 +668,12 @@
     Slick.prototype.changeSlide = function(event, dontAnimate) {
 
         var _ = this,
-            $target = $(event.target),
-            indexOffset, slideOffset, unevenOffset;
+            $target = $(event.currentTarget),
+            indexOffset, slideOffset, unevenOffset, index;
 
         // If target is a link, prevent default action.
         if($target.is('a')) {
             event.preventDefault();
-        }
-
-        // If target is not the <li> element (ie: a child), find the <li>.
-        if(!$target.is('li')) {
-            $target = $target.closest('li');
         }
 
         unevenOffset = (_.slideCount % _.options.slidesToScroll !== 0);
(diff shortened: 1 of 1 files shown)


localpcguy:
I didn't know what minifier you tend to use, or I would have updated the .min.js file as well.

I did do just a tiny amount of cleanup in the same function, moving the if that checks for the 'li' into the switch/case statement since it was only needed for the 'index' case.

localpcguy:
Updated Fiddles:

Bugged JSFiddle:
https://fiddle.jshell.net/localpcguy/vbyg5wxx/3/show/light/
original: http://jsfiddle.net/localpcguy/vbyg5wxx/3/

Fixed JSFiddle:
https://fiddle.jshell.net/localpcguy/vbyg5wxx/4/show/light/
original: http://jsfiddle.net/localpcguy/vbyg5wxx/4/

PR #1529, opened 2015-07-13
Globs in the bower.json main array are not supported

philBrown:
See http://bower.io/docs/creating-packages/#main

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,14 @@
 {
     "name": "slick-carousel",
-    "main": ["slick/slick.min.js", "slick/slick.css", "slick/slick-theme.css", "slick/fonts/*"],
+    "main": [
+        "slick/slick.min.js",
+        "slick/slick.css",
+        "slick/slick-theme.css",
+        "slick/fonts/slick.eot",
+        "slick/fonts/slick.svg",
+        "slick/fonts/slick.ttf",
+        "slick/fonts/slick.woff"
+    ],
     "version": "1.5.6",
     "homepage": "https://github.com/kenwheeler/slick",
     "authors": [


lepipele:
@kenwheeler Can you please merge this to help solve the issue with bower-installer. 

Here is Stackoverflow thread with more info:
http://stackoverflow.com/questions/31351375/how-to-change-bower-installer-path-for-one-component

simeydotme:
I won't merge this yet, as the resource you've linked from Bower states no Font/Image files as `main` .. if you can clean up the PR to accomodate this, I will merge :)

lepipele:
@simeydotme On my machine there is no fonts/images - there is only 1 extra file, namely slick/ajax-loader.gif. So, from what I understand main in bower.json should be:

    "main": [
        "slick/slick.min.js",
        "slick/slick.css",
        "slick/slick-theme.css",
        "slick/ajax-loader.gif",
        "slick/fonts/slick.eot",
        "slick/fonts/slick.svg",
        "slick/fonts/slick.ttf",
        "slick/fonts/slick.woff"
    ]

Correct?

simeydotme:
from www.bower.io:
> __Image and font files__ may be used or referenced within the JS or Sass files, but __are not main files__ as they are not entry-points.

> - Use source files with module exports and imports over pre-built distribution files.
> - __Do not include minified files__.
> - __Do not include assets files like images, fonts__, audio, or video.
> - Filenames should not be versioned (Bad: package.1.1.0.js; Good: package.js).
> - __Globs like js/*.js are not allowed__.

Currently the highlighted parts are wrong, there's no point me accepting a PR which doesn't fix the thing completely :D -- if you can fix the PR to cover all those points, I will merge, else give me some time to do it myself.

Si.

leevigraham:
For future reference: https://github.com/bower/bower.json-spec/issues/47

PR #1598, opened 2015-07-30
Added lazyLoad option 'anticipated' to allow preloading of prev. and next images.

Rendez:
### Description
So that the user experience is better, the images should be loaded on demand, but anticipated, meaning, that the slider will always be one step ahead of the user, by pre-loading the 1 next and 1 previous image so that the loading is seamless, without ever making the user aware that the image was loading. On the other hand, this represents an improvement over loading all the images progressively, as the amount of data to load might not be worth causing also a poor experience, specially on mobile devices with slow connections.

### Demo

http://jsfiddle.net/lmerino/zeL1kmo3/

### Process

There are a few edge cases in which lazyLoad === 'anticipated' might not work as expected, they are rooted on the nature of the current implementation, some bugs and limitations:

1. If slidesToScroll > 1, going from index 0 to -2 and index 8 to 10 only contains one `cloned` item, so scrolling 2 shows an empty image during the animation This can be see in line https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L2034, as `slidesToShow` is used where `slidesToScroll` should be the amount of clones to set-up.

2. If slidesToScroll > 1 and the number of images is an odd number. Similar problem to 1.

3. When `fade == true`, the rangeEnd seems to be +1 count too much, this is either related to another bug or an implementation feature I am not aware of.

Credits to @AlaaSarhan for his collaboration bringing this feature to life.


Siilwyn:
This is a really nice addition when it comes to lazy loading. :+1:

Rendez:
It has to be updated, and I'll get to it if @kenwheeler pop-ins in the thread :)

dansullivan86:
@kenwheeler whats the plan of attack on this?  I could really do with this for a current project.  

I have copied the changes into a fork of the lastest master version: https://github.com/Symbios/slick/tree/dansullivan86-lazyload

How is the minified version built?

kenwheeler:
@Rendez rebase it up bruh

Rendez:
Thanks @dansullivan86. Sorry for the late reply I am supposedly on holidays for 2 more days. I can rebase and build next Tuesday :)

matiwinnetou:
why was this never rebased :(

Rendez:
@matiwinnetou I've just rebased here https://github.com/Rendez/slick-1, I stopped working at my previous company but I managed to recover the repo to my personal account. I have ONLY rebase, which means I haven't yet tested out if the changes work or break. I will do so soon, but please feel free to update the JSFiddle and give me feedback, so we do this collaboratively.

Rendez:
@matiwinnetou @kenwheeler @dansullivan86 @Siilwyn here is a new rebased PR. In the new fiddle added on the comments I've verified that the same code I wrote back last year still works as expected in the current version :)

https://github.com/kenwheeler/slick/pull/2456

rafalf:
@kenwheeler  why is that not present in the  CDN-ed version of slick.js - Version: 1.6.0

Issue #1143, opened 2015-03-16
Disable the use of css transforms


rikkertkoppes:
Since `position:fixed` breaks when using css transforms (<https://code.google.com/p/chromium/issues/detail?id=20574>), a feature to disable it altogether would be nice (and live with left / top transitions).

Since there is already `transformsEnabled` internally, this would be a rather minimal adjustment.

Proposed changes:

- add an option `useTransform`, default to false
- change [this line](https://github.com/theapsgroup/slick/blob/master/slick/slick.js#L1654) to 

        _.transformsEnabled = _.options.useTransform && (_.animType !== null && _.animType !== false);

- add a class if transforms are enabled:

        if (_.transformsEnabled) {
            _.$slider.addClass('use-transform');
        }

- only use transforms for that class

        .slick-slider.use-transform .slick-track,
        .slick-slider.use-transform .slick-list {
            -webkit-transform: translate3d(0, 0, 0);
            -moz-transform: translate3d(0, 0, 0);
            -ms-transform: translate3d(0, 0, 0);
            -o-transform: translate3d(0, 0, 0);
            transform: translate3d(0, 0, 0);
        }

simeydotme:
which element is `position: fixed;` ?

electerious:
Looks like this issue has already been fixed: https://github.com/kenwheeler/slick/pull/1632

vlrprbttst:
even though i use usetransform: false, i still get the transform on my slick carousel

PR #1632, opened 2015-08-12
Optional use of transform

### LINKED ISSUE #1631: Embedded iframe w/video on Firefox (Windows.8.1)
Hi,

I'm having a very specific issue with Firefox v. 40.0 on Windows 8.1.
 I'm embedding a video via iframe and using it as an item in slick. The use of css transforms `transform:translate(640px,0px);` causes the video controls to behave weirdly. 

I'm pretty sure that this is a bug with the way Firefox is dealing with transforms/flash, however, I think that the use of transforms should be able to be overridden anyway. 

#1143 
Without the change:
http://plnkr.co/edit/5EX3XWRQTgnBYzwqUrot?p=preview

With the option disabled:
http://plnkr.co/edit/a4ZB7rni7f4QDrDKB83b?p=preview

ref - #1143

marcellscarlett:
Fixes #1631, #1143

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -81,7 +81,7 @@
                 touchMove: true,
                 touchThreshold: 5,
                 useCSS: true,
-                useTransform: true,
+                useTransform: false,
                 variableWidth: false,
                 vertical: false,
                 verticalSwiping: false,
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -81,6 +81,7 @@
                 touchMove: true,
                 touchThreshold: 5,
                 useCSS: true,
+                useTransform: true,
                 variableWidth: false,
                 vertical: false,
                 verticalSwiping: false,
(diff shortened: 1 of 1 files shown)


kenwheeler:
Thanks dude

electerious:
The default of `useTransform` should be `true`. Performance is so much better when using `transform`. Disabling it will lead to a lot of user never knowing that you can turn it on. Looks like there're only a few edge-cases where you really need to turn it off.

SaraSoueidan:
Hey! I'm currently using the Slider in a project and wanted to point out the importance of having `useTransform` enabled by default. As @electerious mentioned, the performance difference is very big and all performance experts recommend using transforms and not position properties to animate elements.

It would also be nice to add the `useTransform` option to the list available on the plugin homepage, because I couldn't find it without Googling which got me to this thread here.

Cheers!

leggomuhgreggo:
Hey @SaraSoueidan `useTransform` is set to `true` by default in the master. New release is coming soon. Thanks!

SaraSoueidan:
Awesome. Cheers!

PR #1633, opened 2015-08-12
Issue 1481

mpowroznik:
Added new configuration param, to define how many pics buffer when lazyload and fade is enabled.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -85,6 +85,7 @@ nextArrow | string (html|jQuery selector) | object (DOM node|jQuery object) | <b
 infinite | boolean | true | Infinite looping
 initialSlide | integer | 0 | Slide to start on
 lazyLoad | string | 'ondemand' | Accepts 'ondemand' or 'progressive' for lazy load technique
+lazyLoadBuffer | Integer | 2 | Define the number of slides to preload when lazyLoad is used.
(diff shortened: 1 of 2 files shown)


kenwheeler:
can we default this to `slidesToShow`?

mpowroznik:
When you set this default to "slidestoShow" slick, will downlad 4 images (3 from 'slidesToShow' and 1 currently displayed). When it is set to '2', it will download 2 + 1 currently displayed (summary 3).

kenwheeler:
What if you have 9 shown?

mpowroznik:
http://codepen.io/mpowroznik/pen/PPqdmO/ - here you have codepen with my version and param lazyLoadBuffer is set to 0 (this will load only this image, which is required to display) - you can see it in  network tab

http://codepen.io/mpowroznik/pen/wKaEmX - here you have codepen, with your library, and it always preload 3 images (included current image) - when you have added fade option.

http://codepen.io/mpowroznik/pen/wKaEjx - here is the same codepan as above, but this time, fade is set to false, an lazyLoad works pefectly.

My version fix this issue, and you can set, how many images do you want to buffer.

kenwheeler:
Not arguing with your version, arguing about setting the default to 2 rather than a dynamic value

Issue #1537, opened 2015-07-15
Scrollbar jump on previous/next slide click in Chrome


zgreen:
Hi there,

First off, thanks for slick, and all your work on it. It's really terrific, and I use it frequently.

After updating to `1.5.6`, it seems that clicking either the "Previous" or "Next" buttons causes the scrollbar to jump in Chrome (43). [Here is a codepen test using the current version](http://codepen.io/zgreen/pen/vOrVwW), and [here is a test using version `1.5.5`](http://codepen.io/zgreen/pen/xGzQbg). This does not appear to be an issue on either Firefox or Safari (I have not tested IE).

(Setting `accessibility: false` corrects this behavior, but that's obviously not ideal. The work done on https://github.com/kenwheeler/slick/pull/1467 is super helpful, and I am all for making slick more accessible.)

My question is, is this the desired behavior? If it is, I'll close the issue. If not, I'm happy to submit a PR, but I don't want to interfere with the accessibility work that's been done.

Thanks again.

leggomuhgreggo:
Interesting. I'm pretty sure this is related to tab focus from #1467. It looks like when the slider height is larger than the window height, the browser jumps to the top of the focused element.

If you set accessibility to false ([example](http://codepen.io/gwst/pen/LVrXZz)) it wont do that, but that's an unfortunate decision to have to make. 

I'm pretty sure that the "focus jump" will only happen when a user clicks a nav element; I tested with auto play, and if I was scrolled way down the page, it wouldn't jump when it scrolled. So, I guess I'm curious what the use case is for a slider that's larger than the window height, where this would be an issue.

@msrafi @simeydotme what do you think? I was looking around to see if there's a way to control the behavior, but [the options I'm seeing](http://stackoverflow.com/questions/20579803/how-do-i-stop-the-browser-from-jumping-to-the-focused-field) don't look very smooth.

Thanks!

zgreen:
I don't think the carousel itself has to be larger than the window height; there just has to be enough content in the DOM to overflow the window height. [See this fork of my first example](http://codepen.io/zgreen/pen/zGaMRw).

And yes, you're correct, that this appears only to be an issue on nav element clicks. Arrow key navigation does not cause any scrollbar jump.

leggomuhgreggo:
Yeah, it's a little abrupt, but maybe not too tremendously onerous, just because if someone clicked it, they're at least signaling intent to interact with the slider.

dsmy:
[...]

PR #1637, opened 2015-08-13
Fix for #1537 - Scroll jump

msrafi:
Fix for #1537 - Scroll jump


kenwheeler:
Thanks @msrafi

msrafi:
As always, thanks to you @kenwheeler

PR #1644, opened 2015-08-15
Allow custom attribute name for LazyLoad

odie2:
Avoids conflicts with other LazyLoad scripts.

I am using LazyLoad that works globally on website, however I can't exclude slider from simple reason - exclusion's class working only if there are no `data-lazy`. However it is. Then my LazyLoad broking slider for example for fourth item (when 1, 1, centerMode).

I added simple settings so I could just use `data-lazy2`.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -64,6 +64,7 @@
                 infinite: true,
                 initialSlide: 0,
                 lazyLoad: 'ondemand',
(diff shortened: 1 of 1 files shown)


kenwheeler:
how about a namespace. like `data-sl-lazy`

matiwinnetou:
I would like this feature, could be be rebased and merged in?

PR #1650, opened 2015-08-18
Force sliding behavior when centerMode: true and infinite: false, even when count of slides is less than slidesToShow

### LINKED ISSUE #1630: Force sliding behavior when centerMode: true and infinite: false, even when count of slides is less than slidesToShow
The current behavior of slick.js is that, when using centerMode, if there are less slides than the number of slides to show, slick.js will just show all the slides and prevent any sliding. This is not always the intended behavior. I have an application where I always want sliding, not matter how many slides. So even if I show 3 slides at a time and use centerMode, it should start with no slide on the left, the current (and first slide) at the center, and the second slide on the right, instead of showing all slides. Then I should be able to slide to the second and third slides like I would if I had more slides.

It would be handy to have a boolean setting for this behavior, or simply to decide to force the sliding behavior when using centerMode: true, infinite: false, no matter how many slides there are.

This is probably a duplicate of https://github.com/kenwheeler/slick/issues/1549, https://github.com/kenwheeler/slick/issues/1098, https://github.com/kenwheeler/slick/issues/1504, https://github.com/kenwheeler/slick/issues/1302, https://github.com/kenwheeler/slick/issues/262. It seems like a lot of people are having this issue that could be fixed with a new setting.

The workaround I'm currently using is to dynamically set `slidesToShow: 1` when I have 1, 2 or 3 slides, and `slidesToShow: 3` when I have 4 or more slides, but that means the interface looks slightly different depending on the number of slides. I'd like it to be consistent, no matter how many slides there are.

blaiz:
Attempting to fix #1630 without being disruptive.

See this jsfiddle with centerMode: true, infinite: false without the new setting (which seems broken to me):
http://jsfiddle.net/blaiz/bzq3dy5q/

Now this jsfiddle with centerMode: true, infinite: false, with the new setting:
http://jsfiddle.net/blaiz/qLagevqf/

Now this jsfiddle with no centerMode, infinite: true, with the new setting (which doesn't really do anything here):
http://jsfiddle.net/blaiz/2nksdnsv/

To me, the setting with centerMode: true, infinite: false, slidesToShow: 3 with only 3 or less slides seems broken. I added a new setting so that the existing behavior wouldn't change but it seems to me that we could just fix this particular case by adding exception in all the spots where I added `_.options.alwaysSlide`. We could do a check for `_.options.centerMode && _.options.infinite` instead of checking for the `alwaysSlide` setting. Does that make sense?

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -64,6 +64,7 @@ Example:
 Option | Type | Default | Description
 ------ | ---- | ------- | -----------
 accessibility | boolean | true | Enables tabbing and arrow key navigation
+alwaysSlide | boolean | false | Force the sliding behavior even when the count of slides is less than slidesToShow
 autoplay | boolean | false | Enables auto play of slides
 autoplaySpeed | int  | 3000 | Auto play change interval
 centerMode | boolean | false | Enables centered view with partial prev/next slides. Use with odd numbered slidesToShow counts.
--- a/README.markdown
+++ b/README.markdown
@@ -64,7 +64,6 @@ Example:
 Option | Type | Default | Description
 ------ | ---- | ------- | -----------
 accessibility | boolean | true | Enables tabbing and arrow key navigation
-alwaysSlide | boolean | false | Force the sliding behavior even when the count of slides is less than slidesToShow
 autoplay | boolean | false | Enables auto play of slides
 autoplaySpeed | int  | 3000 | Auto play change interval
 centerMode | boolean | false | Enables centered view with partial prev/next slides. Use with odd numbered slidesToShow counts.
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -40,6 +40,7 @@
             _.defaults = {
                 accessibility: true,
                 adaptiveHeight: false,
+                alwaysSlide: false,
                 appendArrows: $(element),
                 appendDots: $(element),
                 arrows: true,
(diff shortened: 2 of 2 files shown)


scriby:
+1

kenwheeler:
The first two examples look the same to me?

blaiz:
@kenwheeler yeah sorry I didn't update the text. I pushed a commit where I actually just fixed what I think is a bug, so instead of using a new setting, I simple force the slider to slide when using centerMode: true and infinite: false.

kenwheeler:
You sir , are correct. Let me give this the old testaroo and I'll get it merged.

blaiz:
I can rebase and squash all 3 commits into one if you want (or can GitHub do that automatically?).

kenwheeler:
I got it

blaiz:
Thanks

gkontokotsios:
@kenwheeler: Hello,

Do you maybe have a plan regarding a release with this commit. I would really like to use the latest version of this plugin along with this fix on my project.

Also thanks for this great slider!

Regards,
Georgios

blaiz:
Hi @kenwheeler, do you still plan on getting this merged?

blaiz:
@kenwheeler any news on this?

blaiz:
This is clearly not going to happen and I'm not even sure that this is still an issue with the latest version so I'll close this PR.

PR #1660, opened 2015-08-21
Fix for responsive breakpoints to work with rows

Bartlomiejsz:
In reference to http://jsfiddle.net/fmo50w7n/47/ error, now it's working correctly :)


odie2:
Doesn't work for me with Opera Developer 33.
When resizing under 500px width, the slides are just destroyed, the same if I run on breakpoint:
image

http://jsfiddle.net/fmo50w7n/61/
(Updated external resources for your last commit from my test site.)

Bartlomiejsz:
Have you tried it with my second pull? Try it and add 'activateRows: true' option, with breakpoint using only 1 row

odie2:
Something is still incorrect.

Init and resize with 149px width:
image

For wider (421px) it's ok:
image

I think that is bug, because plain 2-rows slider works "correct" on 149px:
image

But let me know if I am wrong.

Bartlomiejsz:
Do you mean that all slides are on each other? Because it doesn't have anything to do with slick, but with css. Add overflow: hidden to every slide and it will work:
http://jsfiddle.net/fmo50w7n/64/
If you have another problem, then I'm sorry, i didn't understood your question :)

odie2:
Yes, I saw it in __edit__ already ;)

For me PR is now complete and ready to pull.
Good job.

PR #1682, opened 2015-08-30
Fix outline around main slick list showing up in various browsers

rgbjoy:
slick-list:focus not working to remove the small outline around div. This should fix it.

### DIFF
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -30,9 +30,12 @@
     margin: 0;
     padding: 0;
 }
-.slick-list:focus
+.slick-slide:focus
 {
-    outline: none;
+    outline: 0;
+    -webkit-box-shadow: none !important;
+    -moz-box-shadow: none !important;
+    box-shadow: none !important;
 }
 .slick-list.dragging
 {


guidoprincess:
This would break all accessibility functionality.

rgbjoy:
@guidoprincess The visual highlight is used by screen readers?

odie2:
@rgbjoy The visual highlight is used by almost blind people and some keyboard's people.
Also looks nice in modern browsers if you are using default `outline` style with custom or inherited `outline-color` property.

rgbjoy:
I'm confused. For a almost blind person, what is the highlight around a slider do? Do they want to know they can interact with it?

odie2:
Hmm I am not sure what are talking right know.

Maybe you would check my website for 2 examples of slick:
http://zschocianow.pl/

On top of page is main slider when is no so important and seems like broken with `transform` property, however on bottom of site are partners' logo - there focus is very important as `variableWidth` value is used.

I suggest you rather disabling focus for specific sliders like main slider on home.

PR #1714, opened 2015-09-11
Update README.markdown to include option useTransform

marcellscarlett:
#1632


ryanwalters:
Seems like an important addition to the docs... +1

electerious:
The `useTransform` option is missing on the [official website](http://kenwheeler.github.io/slick/).

PR #1723, opened 2015-09-14
Updated styles to handle dir attribute on multiple elements.

kevinrodrigues:


btpoe:
What are you hoping to accomplish? The `*` is implied when there is not an explicit tag.

Issue #1443, opened 2015-06-11
Breaks on single image in center mode


meggers:
It appears as though Slick breaks on a single image in center mode, in that it doesn't center the single image. Here is a screenshot of the initialized Slick container, demonstrating the issue:

![slick-init](https://cloud.githubusercontent.com/assets/2888003/8117485/0c4f7c92-104e-11e5-9bde-72f9dfc3b580.png)

Here is my setup:

```javascript
      $carousel.slick({
        dots: true,
        infinite: true,
        speed: 500,
        slidesToShow: 1,
        centerMode: true,
        variableWidth: true
      });
```

Has anyone else experienced this?

ejwaibel:
+1 I've had this issue as well. Had to add Javascript to check if there is only 1 item and not enable centerMode.

kenwheeler:
Yeah slick doesn't initialize unless there is enough slides to actually slide.

kushalcoder:
variableWidth: true,
 variableHeight: true


use this. I'will work

PR #1726, opened 2015-09-14
fixed bug when infinite, variableWidth and centerMode are all enabled…

btpoe:
… and the slideCount is less than slidesToShow

This fiddle illustrates the problem:
http://jsfiddle.net/uL6wy4hn/

And this one shows the behavior after applying my changes.
http://jsfiddle.net/5615vLw6/

It's worth mentioning that the conditional I've changed is now almost identical to the one about 6 lines above it. I think this one might have been missed when making the previous change.

This fix resolves issue #1443


kenwheeler:
This is great, thank you

PR #1738, opened 2015-09-16
Arrow Positioning

leggomuhgreggo:
Wanted to propose this as a possible alternative to the current vertical positioning method.

Instead of using a negative margin-top with a value of half the arrow height, I thought it might be better to use transform translate with a y value of -50%; that way it's always centered no matter the height.

Knowing that IE8 support is a goal, I added an IE8 fallback of margin-top: -10px\9;

Less sometimes chokes on \9 so I added LESS styles with the fix for that as a bonus.

[CSS CodePen](http://codepen.io/leggomuhgreggo/full/YywmZz/)
[SCSS CodePen](http://codepen.io/leggomuhgreggo/full/jbWgma/)
[LESS CodePen](http://codepen.io/leggomuhgreggo/full/jbWjjB/)

### DIFF
--- a/slick/slick-theme.css
+++ b/slick/slick-theme.css
@@ -29,8 +29,11 @@
 
     width: 20px;
     height: 20px;
-    margin-top: -10px;
     padding: 0;
+    margin-top: -10px\9; /*lte IE 8*/
+    -webkit-transform: translate(0, -50%);
+    -ms-transform: translate(0, -50%);
+    transform: translate(0, -50%);
 
     cursor: pointer;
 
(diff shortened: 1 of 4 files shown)


kenwheeler:
Beautiful PR. can you update the scss too?

leggomuhgreggo:
Thanks!

Dang, I thought I'd hit the scss in the second commit [here](https://github.com/kenwheeler/slick/pull/1738#commits-pushed-abf7eeb), but I'll admit to not being 100% sure of my PR footing. Let me know if that doesn't fly

PR #1746, opened 2015-09-18
Wrong less syntax

laurent-bettin:
Hi,

In think there are some error in less file. The way you use ampersand generate the wrong rules eg :
.slick-slide.slick-initialized instead of .slick-initialized .slick-slide.

### DIFF
--- a/slick/slick.less
+++ b/slick/slick.less
@@ -56,7 +56,7 @@
         clear: both;
     }
 
-    &.slick-loading {
+    .slick-loading & {
         visibility: hidden;
     }
 }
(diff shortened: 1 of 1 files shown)


kenwheeler:
@leggomuhgreggo is this right? I don't work with less like that

leggomuhgreggo:
Oooo, might be. I'm on it.

leggomuhgreggo:
Thanks for catching that @laurent-bettin 

Fixed in #1747

laurent-bettin:
:+1:

PR #1763, opened 2015-09-24
Make sass file more configurable

Jako:
It would be nice to be able to change the arrow font-size and arrow font-weight by variables.


simeydotme:
not really essential, sorry

simeydotme:
adding "more configuration" as a consideration

Jako:
That way the default SCSS file is not usable and has to be duplicated, when a different font than slick is used. A lot of things in the theme file are configurable. Why not the font size and the font weight.

rudzboy:
+1

PR #1771, opened 2015-09-25
Added arrowsAroundDots option

Jako:
The arrows will be inserted before and after the dots. I have introduced a dotsClass wrapper div to position the arrows and the list inline. Also the arrows are wrapped to retain the toggled display block option (maybe not necessary). An example SCSS for the option is added to the PR.

Thanks for the nice carousel.

### DIFF
--- a/slick/slick-theme.scss
+++ b/slick/slick-theme.scss
@@ -131,6 +131,28 @@ $slick-opacity-not-active: 0.25 !default;
     margin-bottom: 30px;
 }
 
+.slick-dots-wrapper {
+    position: absolute;
+    bottom: -45px;
+    display: block;
+    text-align: center;
+    width: 100%;
+    .slick-prev-wrap, .slick-next-wrap {
+        display: inline-block;
+    }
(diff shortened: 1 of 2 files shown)


gausarts:
Sorry if I missed your point, but isn't this already possible with custom arrows (appendArrows) and appendDots option?
http://jsfiddle.net/o4toq31q/128/

Jako:
Sorry, I did not notice that option. It would have made the patch a bit easier.

It is not the same. The markup on the fiddle makes it impossible to have the arrows displayed **inline** before and after the dots without knowing how many dots are displayed like in this screenshot:

sceen

If there are less dots, the arrows move more in the middle. The prev button has to be placed before and the next button has to be displayed after the dots and all three elements have to be placed inline. It is not possible with display absolute of the arrows.

Maybe an **wrapArrows** option could do that without adding the dotsClass wrapper and the arrow wrappers. If there is a chance of adopting this, I would prepare another pull request.

PR #1776, opened 2015-09-28
Added option to skip specific slides during rotation

skotz:
Added a new option (skipSlideIndexes) which allows you to specify an array of slide indexes to skip over or exclude from becoming the central, active slide.

New functionality: http://jsfiddle.net/6sq969w8/
With infinite scrolling: http://jsfiddle.net/uj0112rm/1/
Backwards compatible: http://jsfiddle.net/bq3hjybv/1/

Works well with multiple slidesToShow and slidesToScroll. Set swipeToSlide to true if you want it to work for mobile touch events.

Tested in Chrome 45, FF 41, and IE 11

BTW... Thanks for the awesome carousel.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -90,6 +90,7 @@ pauseOnDotsHover | boolean | false | Pauses autoplay when a dot is hovered
 respondTo | string | 'window' | Width that responsive object responds to. Can be 'window', 'slider' or 'min' (the smaller of the two).
 responsive | object | null | Object containing breakpoints and settings objects (see demo). Enables settings sets at given screen width. Set settings to "unslick" instead of an object to disable slick at a given breakpoint.
 rows | int | 1 | Setting this to more than 1 initializes grid mode. Use slidesPerRow to set how many slides should be in each row.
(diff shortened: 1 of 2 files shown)


leggomuhgreggo:
Interesting. I'm curious about the use case for this and how it differs from the slickFilter method

skotz:
The slickFilter method removes the slides completely whereas the skipSlideIndexes just pass over them in the rotation without re-initializing the control.

My specific need is to have 3 slides shown at the same time, but to not allow you to "center" on the first or last slides (yet still have them visible). The "new functionality" jsFiddle above shows exactly how I personally intend to use it. If you remove the skipSlideIndexes from that sample, you can scroll to the first/last slide and leave a blank space; that's what I don't want.

leggomuhgreggo:
Ohhh I see. Interesting. Thanks

skotz:
Squashed

photoman:
Has this been included yet?  It's a great idea!

Sproge-Kostya:
Will someone finish this functionality? badly needed

PR #1781, opened 2015-09-29
Update to slick.css to fix a bug

justin-rosecrans:
Found a bug when clicking the prev and next arrows it highlights a blue outline around the last slide within a slideshow or autoscroll.
This is not noticeable with a blue background, however with a white background it is very noticeable.

This bug effects every carousel/slideshow/autoscroll.

This is also noticeable on upon removing the blue background:
http://kenwheeler.github.io/slick/

This change will fix it with "outline-width: 0;" on the slide.
screen shot 2015-09-29 at 11 47 22 am

screen shot 2015-09-29 at 11 47 53 am

screen shot 2015-09-29 at 11 46 47 am

### DIFF
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -81,6 +81,7 @@
 
     height: 100%;
     min-height: 1px;
+    outline-width: 0;
 }
 [dir='rtl'] .slick-slide
 {


christiananguiano:
Came here to report this issue. Adding `outline: 0;` to `.slick-slide` fixes it as well.

tuttleb:
The focus highlighting is an accessibility feature that helps with selecting using the tab key ([see this stack overflow answer for more info](http://stackoverflow.com/a/2943582/2354849)). While this is a decent work-around, I do not think it should be included in the core css. The `accessibility: false` option will also stop the highlighting (but you will also lose tab and arrow key navigation).

justin-rosecrans:
@tuttleb Ah, okay that makes sense. Still it might be better to have the ability to choose which accessibility you want or need.

PR #1791, opened 2015-10-02
Make prev & next buttons work with touch displays

benediktr:
Hey Ken,

I had issues, clicking on the next and prev buttons on mobile devices.
(Since, they only track _click_ events)

What do you think of also tracking the _touchstart_ events?

Best regards,
Benedikt


benediktr:
This was an emulator problem. Not relevant anymore...

PR #1794, opened 2015-10-03
Centralizing Slides

leandrocunha:
centralizing-slides

Added margin auto on slick-track, to center slides if no have a minimum in the cases of dynamic slides

### DIFF
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -55,6 +55,7 @@
     position: relative;
     top: 0;
     left: 0;
+    margin: 0 auto;
 
     display: block;
 }
--- a/slick/slick.less
+++ b/slick/slick.less
@@ -43,6 +43,7 @@
 .slick-track {
     position: relative;
     left: 0;
+    margin: 0 auto;
     top: 0;
     display: block;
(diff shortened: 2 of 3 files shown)


kenwheeler:
This might be better as just a class that is added so that people can implement the not enough slides styles how they'd like to, as opposed to implicit centering

leandrocunha:
of course, but as in my test this feature had no side effect I think this
behavior could be native

2015-10-03 13:25 GMT-03:00 Ken Wheeler <notifications@github.com>:

> This might be better as just a class that is added so that people can
> implement the not enough slides styles how they'd like to, as opposed to
> implicit centering
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/kenwheeler/slick/pull/1794#issuecomment-145262901>.
>



-- 
[image: http://frango.herokuapp.com] <http://frango.herokuapp.com>
  *Leandro Cunha*Empreendedor motivado Front-End ExpertBack-End autodidata
*http://frango.herokuapp.com* <http://frango.herokuapp.com/>

PR #1823, opened 2015-10-16
Update README. (Fix Settings table)

ByScripts:
There was some pipe breaking the table layout, and `<button>`'s code was not escaped.


altern8tif:
:+1: 

Markdown syntax needs to be [Github-flavored](https://help.github.com/articles/github-flavored-markdown/) to display properly.

kenwheeler:
Thanks!

Issue #1838, opened 2015-10-22
Missing dots if using multiple slides to show


ethanclevenger91:
http://codepen.io/ethanclevenger91/pen/xwYERX

You'll notice there are 7 slides, but only 5 dots. If you use the arrows to get to slides 6 or 7, no dot is highlighted. Seems unlikely this is intentional. I'll work on a patch unless you've got something in mind.

janh92:
Will this be fixed soon?

kenwheeler:
Yes

codysatellite:
Any progress on this?

codysatellite:
To anyone it may concern, this issue seems to be fixed on version 1.5.9. Thanks Ken.

PR #1840, opened 2015-10-22
Dots fix on slidesToShow

ethanclevenger91:
Here's broken again: http://codepen.io/ethanclevenger91/pen/xwYERX/

Here's fixed: http://codepen.io/ethanclevenger91/pen/WQMGaw

In `getDotCount()` I swapped the breakpoint increment to `slidesToScroll` rather than `slidesToShow`.

Unrelated: I noticed in the master branch, `slick.js` (updated < 1 month ago) causes the slide animation to look strange. Draggy? Gross? Either way, not as clean as the `slick.min.js` being served up by CDNjs, v1.5.8. I think a change may have been made that affected this. I edited and fixed both in my branch, so if you look at the fixed pen, it's currently set to load my new `slick.min.js` based on the CDNjs v1.5.8 (which is probably the same as master). If you go into the JS resources and swap `dots-fix/slick/slick.min.js` for `dots-fix/slick/slick.js`, you'll see the draggy animation I'm talking about. I can roll back my changes to `slick.min.js` before merging if you want.


ethanclevenger91:
Oh, fixed #1838

ethanclevenger91:
Figured out the funny animation, related to https://github.com/kenwheeler/slick/pull/1632 - needed to set `useTransform:true`. Rolling back changes to min.

ahmadalfy:
Looks good to me! Let's wait for @kenwheeler

miscbits:
http://codepen.io/anon/pen/yYrOGP

Haven't checked if this is related to the fix or if this is already an issue in slick, but when slides to scroll is less than slides to show (here its 2) then there is one too many dots.

ethanclevenger91:
Hm, it's not even necessarily that there are too many dots - for the third dot, the slides don't respond, but the last dot scrolls _3_ slides instead of 2.

PR #1846, opened 2015-10-29
Change .finds to .children to support nested slicks

jkbailey:
You can see the branch working here: http://jsfiddle.net/fmo50w7n/125/
You can see the current release working here: http://jsfiddle.net/fmo50w7n/127/

To see the animation break due to slick-center not being applied correctly, navigation from slide 1 to slide 6

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1140,7 +1140,7 @@
         centerOffset = _.options.centerMode === true ? _.slideWidth * Math.floor(_.options.slidesToShow / 2) : 0;
 
         if (_.options.swipeToSlide === true) {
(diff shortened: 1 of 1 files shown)


ahmadalfy:
Looks good to go! I faced that problem and fixed it by destorying and reinitalizing the nested slider everytime the parent change. This will fix it.

Can you provide a jsfiddle with an example so that @kenwheeler can have a quick look at it working?

jkbailey:
Thanks @ahmadalfy

daversm:
@jkbailey I'm having an issue where a swipe gesture on my nested slick slider also swipes the parent slick slider. I wanted it so that the parent won't slide if I swipe left/right on the child. Will your changes also fix that issue ?

PR #1882, opened 2015-11-11
changed return value of getDotCount 

miscbits:
to account for number of slides showing compared to the number of slides scrolled to fix the dot count bug

I haven't tested extensively, but this works for every case where slidesToShow is greater than slidesToScroll. It should work the reverse as well.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1011,7 +1011,7 @@
             }
         }
 
-        return pagerQty - 1;
+        return pagerQty + _.options.slidesToShow - _.options.slidesToScroll - 1;
 
     };
 


miscbits:
sorry had no jsFiddle up.
broken: http://jsfiddle.net/jjzbmdbq/
fixed: http://jsfiddle.net/ua20L7yo/

ahmadalfy:
This fix is confusing. I thought your example was broken till I saw the original slider with the broken behavior. @kenwheeler what do you think?

miscbits:
@ahmadalfy what part needs more clarification? 

The incorrect number of dots appears when the number of slides to show is greater than the number of slides to scroll.

ethanclevenger91:
Hrm, is this related to #1840 ?

miscbits:
yeah. After looking at his fix its the same as mine, just done a different way.

miscbits:
Closing this due to inactivity.

PR #1886, opened 2015-11-12
Fix bug with "variableWidth: true" and and "rtl: true"

pazaricha:
@kenwheeler First of, Thanks for an awesome plugin :smile: 

Bug reproduced: http://jsfiddle.net/pazaricha/qy18dm5k/
Bug fixed: http://jsfiddle.net/pazaricha/7uLwo1hu/

This works both with the following settings: 
```javascript
centerMode: true || false
infinite: true || false
```
`slidesToShow & slidesToScroll` are irrelevant as far as I understand when `variableWidth` is set to `true`. (correct me if I am wrong on this)

I'd live to hear your comments and also if you'd like me to bump the version number across the files I will be happy to do so.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1075,7 +1075,15 @@
                 targetSlide = _.$slideTrack.children('.slick-slide').eq(slideIndex + _.options.slidesToShow);
             }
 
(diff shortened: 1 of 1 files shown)


pazaricha:
Links we're broken they are now fixed.

PR #1904, opened 2015-11-19
Update README.markdown

aicarlson:
Removed what appeared to be a typo on line 94.


ahmadalfy:
Thanks for the notice

PR #1931, opened 2015-12-01
Whitish effect when slides are switched

chestozo:
Hello )
Thank you for a great job.
This slider plugin in really nice one.
And also your code is organised really well!

I have an issue with switching slides:
when it happens in my case — I see a whitish effect.
whitish

This happens maybe because images in my case have almost the same background.
So that it is easier to see that these images are similar.

This effect is happening because we start to fade out old slide and fade in new slide at the same time https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L2209.

A simple fix could be to move `_.fadeSlideOut(oldSlide);` inside `_.fadeSlide` callback so that we start to fade out old slide when new one is already fully visible.
This looks better:
ok

Is there any chance that this kind of pull request will be accepted? :)
Thank you

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -60,6 +60,7 @@
                 easing: 'linear',
                 edgeFriction: 0.35,
                 fade: false,
+                fadeInFirst: false,
                 focusOnSelect: false,
                 infinite: true,
                 initialSlide: 0,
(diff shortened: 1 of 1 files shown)


chestozo:
Yes please )

chestozo:
This is how it looks with updated script
http://jsbin.com/soselidaye/1/edit?html,output

There is a problem with the old slide been seen under the new one.
But if slides are of the same size - this will not be visible.

chestozo:
Still there is now this jumping effect at the bottom (if slides are of different size):
bottom

This is how it looks in current version
bottomd

chestozo:
Any plans for this PR?

PR #1941, opened 2015-12-03
Added note about browser support

ahmadalfy:
We are almost in 2016 and people are still asking for it ...


simeydotme:
thanks @ahmadalfy

PR #1946, opened 2015-12-07
Add autoplay stop method.

extempl:


ahmadalfy:
Aren't there any other method to stop the slider? pinging @kenwheeler

kenwheeler:
@ahmadalfy `slick('pause')` stops it. `slick('setOption', 'autoplay', false)` kills the option

PR #1948, opened 2015-12-07
Fix touch event

tomaszkepa:


ygrenzinger:
Hello,

If I understand correctly, your code fixe the touch event on the slideshow ?

Thanks

PR #1960, opened 2015-12-10
Use variable paths in LESS.

adjohnson916:


adjohnson916:
The Travis build failed due to it detecting this as a Rails/Rake app. There should probably be a Travis config specifying the expected language, probably Node, though I don't see any tests anyway.

rockfreak:
Test results if it's of any use for you. I too have made these changes to the less file and would like to see them pulled into the master.

https://travis-ci.org/kenwheeler/slick/builds/96163577

PR #1981, opened 2015-12-21
Update slick-theme.less with variables for font path and image loader path

aicarlson:
Adds variables that were already declared for the font and loader paths


aicarlson:
I'm not very familiar with Travis CI so I'm not sure why the build is failing - Any input? I'm happy to fix things but it looks like a rakefile is missing?

rocknruby:
+1

PR #1994, opened 2015-12-30
New stylus implementation

Jorgewlf88:
Stylus implementation, it's cleaned


simeydotme:
thanks @Jorgewlf88 , but we need to discuss whether having source files for multiple pre-processors is relevant anymore. If you leave your branch alive, we will still be able to pull from it in the future.

:bow: 

Thanks for your effort!

PR #2001, opened 2016-01-05
[lazyLoading] Fix partiallly visible elements

jzastrow:
To pay more attention to the scroller we want to display 60% of the last picture only. We set ```slidesToShow=4.6```. Without this patch the 4.6 value is used to slice the array which results in the ```4``` beeing used.
This patch uses the ```Match.ceil()``` to fix this.


kenwheeler:
Thanks

PR #2006, opened 2016-01-09
update slick-theme.css to avoid erro in gulp

sankaran8:
Changed margin-top: -10px/9; /*lte IE 8*/ to margin-top: -10px; /*lte IE 8*/

It was causing issue when included in gulp.

### DIFF
--- a/slick/slick-theme.css
+++ b/slick/slick-theme.css
@@ -30,7 +30,7 @@
     width: 20px;
     height: 20px;
     padding: 0;
-    margin-top: -10px\9; /*lte IE 8*/
+    margin-top: -10px; /*lte IE 8*/
     -webkit-transform: translate(0, -50%);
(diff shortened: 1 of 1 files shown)


kenwheeler:
Thanks!

leggomuhgreggo:
Just noticed this! A while ago I [submitted a PR](https://github.com/kenwheeler/slick/pull/1738) to switch to using transform for vertical centering, because it doesn't require a fixed pixel offset of half the height of the arrow. 

Having the /9 following the margin top value makes it so that the style only registers for IE8, which doesn't support transform. I thought it'd be good to have as a fallback for the older browser, but, given that i causes problems for gulp and is tricky in other regards, perhaps removing the property entirely is the right way to go, and IE8 users can have a slightly off centered arrow.

Alternatively we could revert to the fixed value.

Thoughts?

PR #2008, opened 2016-01-10
make overriding sass variables possible when used with compass

patchee500:
When overriding the sass variables "$slick-loader-path" and "$slick-font-path" the resource paths will not actually change because slick-theme.scss will call compass function image-url() even if the variable has been overridden. This basically makes it impossible to override those variables when used with compass, as the function image-url() will always exist and if not specified, return some default value.

This fixes that. If the variable has been set before, it will take the overridden value. If it has not been set, it will check the existence of the image-url() function and use that if it exists, otherwise it will default to the original values.

Also see this SO Question: http://stackoverflow.com/questions/34572876/is-it-possible-to-not-use-turn-off-compass-url-helpers-on-purpose

### DIFF
--- a/slick/slick-theme.scss
+++ b/slick/slick-theme.scss
@@ -26,7 +26,7 @@ $slick-opacity-not-active: 0.25 !default;
         @return image-url($url);
     }
     @else {
-        @return url($slick-font-path-default + $url);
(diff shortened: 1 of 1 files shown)


patchee500:
noticed there is still an issue in the code. gonna fix that.

PR #2009, opened 2016-01-10
Added setting to stop event propagation and allow nested carousels

juancamiloestela:
I need to have nested carousels and allow the user to swipe without moving the parent carousel. I've added the ability to stop the user input event from propagating to the parent carousel via a setting variable.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2333,6 +2333,10 @@
 
         var _ = this;
 
+		if (_.options.stopSwipePropagation){
+			event.stopPropagation();
+		}
+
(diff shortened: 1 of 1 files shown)


simeydotme:
there's a few problems with this PR.
- code style
- no default set
- no fiddle to demonstrate

but not only that, it seems like a strange use-case... I wonder if we are looking to support nested carousels at all? I'll put it in for consideration in 2.0 :)

nicooga:
When on mobile or any touch capable devices, swiping on the carousel causes the sidenav to appear. I think this can be a common problem. Is there something we can do to get this merged or work around the issue?

PR #2026, opened 2016-01-15
Fix keep a reference to each Slick instance

mkdgs:
When you have multiple slider on the same page there no way to get access the instance for each of them 
So storing an instance with the element fix this issue.

that works with this fix:
```
 $('.my-slick-slider-element').each(function() {
            var slick = $(this).data('slickInstance');
            slick.slickSetOption('arrows',false);
});
```

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2658,8 +2658,10 @@
             i,
             ret;
         for (i = 0; i < l; i++) {
-            if (typeof opt == 'object' || typeof opt == 'undefined')
+            if (typeof opt == 'object' || typeof opt == 'undefined') {
                 _[i].slick = new Slick(_[i], opt);
+                $(_[i]).data('slickInstance', _[i].slick);
+            }
             else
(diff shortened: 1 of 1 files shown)


kenwheeler:
Isn't this already available at $(this).slick ?

kenwheeler:
Or this.slick rather

mkdgs:
$(this).slick work only when you have only one slider, something weird appear with many.
I suppose the reference is overwritten and only the last is defined

kenwheeler:
http://jsfiddle.net/7nmgmf4p/1/

mkdgs:
No, there is a bug
http://jsfiddle.net/7nmgmf4p/2/

see the last $('.slider.no-dots') throw:
TypeError: instances[1] is undefined

kenwheeler:
No, there actually is not. Your selector only selects the second slider, so it would live at instance[0] because their is only one product of the iteration, not two, like I had in my example.

mkdgs:
You're right, forget me... i am going to go sleep a little

PR #2027, opened 2016-01-15
Fix readme default useTransform value

krzysiek1507:


krzysiek1507:
@kenwheeler Can you release new version with `useTransform: true` as default?

kenwheeler:
Yeah, fixing some other stuff first

PR #2028, opened 2016-01-15
fixing bug when "init" event fires before breakpoint is initialized

openmarco:
Hi,

We have a problem with "init" event and option "slidesToShow" on multiple breakpoints.
In short, the "init" event occurs before the checkResponsive() action and final set of clonned elements is diffirent from the "init" moment.

Detailed description on the jsfiddle link:
http://jsfiddle.net/h1u7ohzt/2/

Openmarco team


kenwheeler:
Can you remove the .min?

PR #2030, opened 2016-01-15
fixing bug when "init" event fires before breakpoint is initialized

openmarco:
Hi,

We have a problem with "init" event and option "slidesToShow" on multiple breakpoints.
In short, the "init" event occurs before the checkResponsive() action and final set of clonned elements is diffirent from the "init" moment.

Detailed description on the jsfiddle link:
http://jsfiddle.net/h1u7ohzt/2/

Openmarco team

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -175,7 +175,6 @@
 
             _.registerBreakpoints();
             _.init(true);
-            _.checkResponsive(true);
 
         }
 
@@ -1204,6 +1203,7 @@
             _.initializeEvents();
             _.updateArrows();
             _.updateDots();
+            _.checkResponsive(true);
 
         }
 


simeydotme:
Wow. This deserves a :clap: slowclap! :grin: 
Great fiddle, and great explanation for such a small code change!

Great work!

simeydotme:
I am going to spend a bit of time thoroughly investigating any possibly regressions with this before I merge... but looks good so far.

Stay tuned!

openmarco:
Great! Thank you!

launchoverit:
Just FYI, it looks like there's a similar problem with the aria attributes, it looks like the init event is fired before the attributes are added(?)

http://jsfiddle.net/cn56zoLq/2/

Thanks for your work on this!

PR #2031, opened 2016-01-15
Feature/wordpress vip compatibility

doubleswirve:
Replicating some small updates we made to get WordPress VIP to accept Slick. Changes mainly concern replacing HTML string concatenation with jQuery DOM construction and avoiding the use of jQuery's `.html` method.

Here's JS Fiddle using the changes with the site demos:

https://jsfiddle.net/doubleswirve/21bvhboz/14/

Thanks!

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -790,7 +790,7 @@
         if(_.options.rows > 1) {
             originalSlides = _.$slides.children().children();
             originalSlides.removeAttr('style');
-            _.$slider.html(originalSlides);
+            _.$slider.empty().append(originalSlides);
         }
 
     };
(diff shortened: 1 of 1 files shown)


kenwheeler:
Whats the motivation here regarding acceptance?

doubleswirve:
I think it mostly centers around the possibility of XSS:

https://vip.wordpress.com/2015/03/25/preventing-xss-in-javascript/

Note from VIP:

> I found some html string construction issues in slick carousel, and static/js/feature_rel-post.js in the
> latest commits. The related posts should be easy to refactor, but the slick carousel looks to rely on html
> string construction at an architectural level, and refactoring may not be trivial

After making the above changes, we were able to deploy.

kenwheeler:
Figured as much, thanks, good to go

PR #2037, opened 2016-01-16
fix jumping when infinite is off

krzysiek1507:
Before:
http://jsfiddle.net/co8po9fr/2/
Click fast 5 times `prev`, then `next` and `prev`. Slides are jumping.

After:
http://jsfiddle.net/q1qznouw/319/
Same instruction and everything is ok.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2131,6 +2131,10 @@
 
         sync = sync || false;
 
+        if (_.options.infinite === false && (index < 0 || index >= _.slideCount)) {
+            return;
+        }
+
         if (_.animating === true && _.options.waitForAnimate === true) {
             return;
         }


simeydotme:
Before this is merged, I'd like to see it tested against the `asNavFor` method; specifically the options in this fiddle:
http://jsfiddle.net/xd2h934n/ __<-- example before my fix__
http://jsfiddle.net/6vby3gt9/1/ __<-- example of working__
(second slider in result panel)

I just updated `master` with a fix for this issue and think there could be some conflict... `asNavFor` utilizes the `slideHandler` method, but `slideHandler()` has some problems with all those `return`s happening before the main body... quite a bad practise I think.

So please make sure the `asNavFor` is still working correctly with regards to updating classes and such.

PR #2040, opened 2016-01-17
Fixes #1248

### LINKED ISSUE #1248: The class .slick-active appears on all slides, should only be the first slide
I am using the asNavFor option and all of the items in the nav load with the class .slick-active. Only the first option should be active. It's not until you click, that it begins to work correctly. Is there a way to fix this when it loads.

All items load as slick-active. If you click one of them, the .slick-active class leaves all of the other elements. This is how I want it to load though, with only the first item active.

simeydotme:
Because of a condition in `sliderHandler()` checking if `slidesToShow <=
slides` we could never set the slide classes for a nav slider with this condition.

Had to split the `asNavFor()` method in to two parts, and set the slide classes
in the `slideHandler()` method when `options.asnavfor` is set and the
`slidesToShow <= slides` -- the other option (which I didn't want to tackle) was
to re-architect the whole `slideHandler` method.

bit hacky. but stable. This architecture should be reconsidered in 2.0, @kenwheeler

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -318,14 +318,23 @@
 
     };
 
-    Slick.prototype.asNavFor = function(index) {
-
+    Slick.prototype.getNavTarget = function() {
+        
         var _ = this,
             asNavFor = _.options.asNavFor;
 
         if ( asNavFor && asNavFor !== null ) {
             asNavFor = $(asNavFor).not(_.$slider);
         }
+
+        return asNavFor;
+
+    };
+
+    Slick.prototype.asNavFor = function(index) {
+
+        var _ = this,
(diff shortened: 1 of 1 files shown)


elbsurfer:
Hello,
this issue is still persistent in 1.5.9. I have 3 images which should all be shown at one time. I set up slidesToShow to 3 and when I navigate through the pictures "slick-current" is not set properly.

Devilmore:
We are having the same issue, also with 1.5.9.

simeydotme:
That's because `v1.5.9` is the same version as before the fix, there needs to be a build released :smile:

PR #2044, opened 2016-01-19
Remove unneeded -moz-box-sizing from slick.css

### LINKED ISSUE #1121: -moz-box-sizing: border-box can be removed
Suggest ' -moz-box-sizing: border-box;' to be removed from the CSS now. That vendor prefixed attribute was only required up until Firefox 28 (and we're now on Firefox 36 - and it auto updates so I think no need to support such an old version).

REF: http://caniuse.com/#search=box-sizing

coliff:
Remove unneeded -moz-box-sizing. Fixes #1121

### DIFF
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -4,9 +4,7 @@
     position: relative;
 
     display: block;
-
-    -moz-box-sizing: border-box;
-         box-sizing: border-box;
+    box-sizing: border-box;
 
     -webkit-user-select: none;
        -moz-user-select: none;


leggomuhgreggo:
One of the tedious side effects of supporting CSS compilers is that there're multiple files for styles. Would you be willing to update the slick.less and slick.scss files as well?

coliff:
Cool, done. sorry for hassle of doing separate PRs for each file. wasn't sure how to do a PR with multiple files using just the web interface to Github

leggomuhgreggo:
Thanks for taking the time!

Note: PR #2050 #2049 and #2044 Are all related, and have to do with removing vestigial vendor prefix for box sizing.

Issue #1516, opened 2015-07-08
When Autoplay is enabled pause does not last forever. 


tkiethanom:
When autoplay is set to true, and slickPause was called the carousel would suddenly resume playing again after some time. Maybe a function of the autoplay speed.

simeydotme:
there's no JSFiddle, or example of how to recreate ...
Closing, please read the contribution guidelines when opening tickets :)

Best, Simon.

PR #2047, opened 2016-01-19
Use another variable to track the autoplay state

tamcy:
When autoplay and pauseOnHover are true, it is impossible to really pause the slider because it will always re-register itself to autoplay again when the mouse moves over and away of it. To fix this, we probably need to set autoplay to false when pause is called. However, I don't think it is appropriate to change the value of autoplay internally as this is an option and shouldn't be used to track state. So in this pull request I introduce a new variable "autoplaying" to track the autoplay state.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -126,6 +126,7 @@
             _.cssTransitions = false;
             _.hidden = 'hidden';
             _.paused = false;
+            _.autoPlaying = false;
             _.positionProp = null;
             _.respondTo = null;
             _.rowCount = 1;
@@ -747,7 +748,7 @@
 
             $('li', _.$dots).off('click.slick', _.changeSlide);
 
-            if (_.options.pauseOnDotsHover === true && _.options.autoplay === true) {
+            if (_.options.pauseOnDotsHover === true) {
 
                 $('li', _.$dots)
                     .off('mouseenter.slick', $.proxy(_.setPaused, _, true))
@@ -1249,7 +1250,7 @@
             }, _.changeSlide);
         }
 
-        if (_.options.dots === true && _.options.pauseOnDotsHover === true && _.options.autoplay === true) {
+        if (_.options.dots === true && _.options.pauseOnDotsHover === true) {
             $('li', _.$dots)
                 .on('mouseenter.slick', $.proxy(_.setPaused, _, true))
                 .on('mouseleave.slick', $.proxy(_.setPaused, _, false));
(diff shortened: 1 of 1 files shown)


tamcy:
This may be the cause of #1516.

simeydotme:
what's wrong with:

```js
$("button").on("click", function() {

$slider
  .slick("setOption", "autoplay", false)
  .slick("pause");

});
```

:question:

tamcy:
It really depends on how `autoplay` and `pauseOnHover` are supposed to work in this library.

To my understanding, an `autoplay` option means that the carousel should advance itself without calling `play()` manually after the carousel is initialized. Nothing more, nothing less. If this sounds sensible, then this particular `autoplay` option is actually synonym to something like `playOnStart`.

So imagine this `autoplay` option is really named `playOnStart`. It doesn't really make sense to "pause playback on mouse hovers" and "resume playback when mouse exits" because this `playOnStart` option is true.

When `pause()` is called explicitly, I expect the carousel to stop completely unless I call `play()` again. Moving the mouse to and out of the carousel should not trigger this `play()` action again. `pauseOnHover` should only resume playback when such pause is triggered by the same hover action.

To my observation, the root cause is that the meaning or responsibility of `autoplay` is being mixed up.  It is currently being used for two purposes:

1. to determine whether to automatically call `play()` on init (=playOnStart, as an option).
2. to track whether the current carousel is "automatically advanced" (= isPlaying, as a state flag). A carousel can be "playing" or "stopped", so this `autoplay` can also be set to `true` or `false`.

simeydotme:
Great comment, and I totally agree with your reasoning, but I think it can be achieved without adding another (somewhat confusing) internal state tracker... I kind of feel it needs to be refactored completely without destroying the API. -- the current interactions between `pause`/`play`/`mouse` are spaghetti.

Any interest in tearing it down, and starting fresh while maintaining the public api, @tamcy ? :)

tamcy:
Thank you for your invitation, but as I am young to Javascript I don't think I can come up with something better than the current one. After all, proposing a fix and maintaining a library are two different stories. But I still think it is inevitable to introduce a new variable to track the playback state, of course its name can be chosen so that it isn't as confusing as the one I use in this PR. Thanks again.

simeydotme:
no problem @tamcy  :)
and you're right, it was inevitable to add another state tracker ;) -- but I had to re-think the states.
there's `paused`, `interrupted` and `focussed` now, each tracking their own thing.

Sorry I couldn't get your PR merged, but don't worry about being new to JS... keep up the good work, im sure there's other stuff you can contribute to Slick! :bow: :grin:

PR #2049, opened 2016-01-20
Update slick.scss

coliff:
Update slick.scss


leggomuhgreggo:
PR #2050 #2049 and #2044 Are all related, and have to do with removing vestigial vendor prefix for box sizing.

PR #2057, opened 2016-01-22
End of carousel and resize with decimal value for slides to show

earlyster:
when slidesToShow is decimal value ie 3.5 we show 3 and 1/2 of a slide when we are at the end of the carousel and resize we get blank carousel. Also in some cases the order of the slides are not correct.

https://jsfiddle.net/earlyster/q1qznouw/321/

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "slick-carousel",
-  "version": "1.5.9",
+  "version": "1.5.10",
   "description": "the last carousel you'll ever need",
   "main": "slick/slick.js",
   "repository": {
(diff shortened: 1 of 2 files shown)


simeydotme:
@kenwheeler whats the word on floating point `slidesToShow` ? I never even considered this before...

kenwheeler:
I mean, its not officially supported, but even i've used it as a workaround.

earlyster:
Will you accept this pull request?

kenwheeler:
I have to test it thoroughly first

earlyster:
Thanks!

PR #2064, opened 2016-01-25
Add Example Syntax for slickSetOption to Readme

willthemoor:
Tiny update to the readme. I had to experiment until I sorted out the syntax for the methods that take options. Here's me throwing other noobs like me a bone. :)


simeydotme:
thanks :)

PR #2068, opened 2016-01-25
[ bugfix 2055 ]

### LINKED ISSUE #2055: Responsive: Slides disappear on resize
I'm facing the same problem with this issue: https://github.com/kenwheeler/slick/issues/1995 which is already closed.

Here's the fiddle: https://jsfiddle.net/wL1t5qu8/

Step:
- Run the page on larger screen than breakpoint to see that there're 5 items and cannot autoplay because slidesToShow = 5. (correct)
- Resize the page to lower than breakpoint (768) to see that it show 3 items with autoplay. (correct)
- Wait for carousel to autoplay 1 item.
- Resize the page back to original size.
- The item showing is now 4 instead of 5 (bug)

Actually, if you wait for carousel to slide pass X items, when you resize back it's going to show 5 - X item. The only thing that's not reproduce this bug is when you wait carousel to slide pass exactly 5 items.

simeydotme:
- error with the handling on slider refresh
- happens when going to a breakpoint with more slides to show than actual
  slides.
- fixed bad condition structure.

fixes #2055


claudiomedina:
Can you update the npm version to 1.5.10?

kenwheeler:
@simeydotme how are we looking for a release?

PR #1000, opened 2015-02-05
switch keyHandler position in rtl

boynet:
In rtl the 'left' arrow should swipe next and right arrow should swipe prev

PR #2069, opened 2016-01-25
swapping keyboard arrow pressing in rtl

boynet:
in rtl the keyboard left arrow should swipe next and the right arrow should swipe previous see https://github.com/kenwheeler/slick/pull/1000


simeydotme:
http://jsfiddle.net/simeydotme/abjdmujd/1/
I did a fiddle for you.

PR #2074, opened 2016-01-27
set max-width of image in slider

erickLeandro:
- now at lower resolutions the picture will occupy 100% of the screen.


simeydotme:
Please create a Fiddle to demonstrate the behaviour before and after ... I'm dubious whether this belongs in slick's css file :)

Issue #984, opened 2015-02-02
Extra 30px bottom margin problem


jsmad:
The default CSS adds a mandatory 30px bottom margin, which causes all sorts of display problems if the html designer is not aware of it.

It would be better if this margin was optional, and only set when the "dots" are enabled.

```css
/* Dots */
.slick-slider
{
     margin-bottom: 30px; 
}
```

tomaszpoliszuk:
can be easy (i hope;) done:
<pre>
if (_.options.draggable === true) {
	$(_.$slider).addClass('slick-dotted');
}
</pre>
inside
<pre>Slick.prototype.init = function()</pre>
and
<pre>_.$slider.removeClass('slick-dotted');</pre>
inside
<pre>Slick.prototype.destroy</pre>
and of course
<pre>
.slick-dotted {
	.slick-slider {
		margin-bottom: 30px;
	}
}
</pre>
in slick-theme.scss

seriema:
Are you using 1.4? The "pretty" styling stuff is now in `slick-theme.css` instead of `slick.css`. So these things would be easy for your designer to use/override or ignore the theme completely.

tomaszpoliszuk:
Yes, but @jsmad has good point - if you use default slick-theme.css there's no need for margin-bottom
 in case you have dots disabled.

seriema:
:+1:

leggomuhgreggo:
The 'dotted' class addition is definitely the better way to go, but another alternative, which might be useful, would be to replace the margin-bottom on the slick-slider with `.slick-list:nth-child(2):nth-last-child(3){margin-bottom: 30px;`} 

This is pretty fragile tho.

leggomuhgreggo:
Merged this fix :) Closing

PR #2090, opened 2016-02-01
Add slick-dotted to slider when dots enable so spacing is triggered instead of default

leggomuhgreggo:
This PR is in reference to #984. Basically the idea is that the spacing that's added to accommodate dots is still there even if `dots: false`

Wouldn't want to invite a deluge of messy class based styling, but I encounter this somewhat regularly; so, I wanted to submit it for your consideration.

Actually, as I dove into it a bit, I noticed the dots ul had some margin which seemed to make the positioning a bit convoluted, so I hollered at that too so offset essentially equals the height of the dots.

To test, toggle the dots on and off. Chapelle's Show gif to illustrate the spacing:
[SCSS version](http://codepen.io/leggomuhgreggo/pen/jWpMxb?editors=0010)
[LESS version](http://codepen.io/leggomuhgreggo/pen/JGBboa?editors=0010)
[CSS version](http://codepen.io/leggomuhgreggo/pen/eJjBJy?editors=0010)

### DIFF
--- a/slick/slick-theme.css
+++ b/slick/slick-theme.css
@@ -114,7 +114,7 @@
 }
 
 /* Dots */
-.slick-slider
+.slick-dotted.slick-slider
 {
     margin-bottom: 30px;
 }
@@ -122,12 +122,13 @@
 .slick-dots
 {
     position: absolute;
-    bottom: -45px;
+    bottom: -25px;
 
     display: block;
 
     width: 100%;
     padding: 0;
+    margin: 0;
 
     list-style: none;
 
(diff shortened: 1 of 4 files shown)


simeydotme:
looks good, @leggomuhgreggo  :)
do you know how to squash commits in an interactive rebase? -- some of those commits like: "underp a derp" would be better off squashed out :P

leggomuhgreggo:
I was wondering about that! Truthfully, I whipped this through the github web interface. Next time I'll be an adult and use the cl properish. Thanks @simeydotme

simeydotme:
sweet! looks good to go!

PR #2094, opened 2016-02-03
Fix a bug after call slickRemove add slickAdd

GrayYoung:
When use slickRemove to empty a carousel, *e.g while($slick.slick('slickRemove', 0));*, $prevArrow and $nextArrow will set its aria-disabled into true, so when use slickAdd add new items into the carousel, $prevArrow and $nextArrow need to be remove their aria-disabled attributes or set their aria-disabled attributes into false.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -443,8 +443,8 @@
 
             if( _.slideCount > _.options.slidesToShow ) {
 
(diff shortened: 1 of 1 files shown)


simeydotme:
I need to a see a fiddle of this happening, and it being fixed.
I took a quick look through the code, and it should be handled by "updateArrows"

GrayYoung:
@simeydotme, OK, thanks.

PR #1139, opened 2015-03-13
beforeChange event; event.preventDefault() to block slide change

adam-lynch:
Related to #1138

Issue #1138, opened 2015-03-13
Can't prevent change of current slide


adam-lynch:
I'd love to be able to prevent a user going to a certain steps until some conditions are met. 

It's great that there are so many events but we need to be able to call `e.preventDefault()` or something like that. Or even if in a `beforeChange` callback you could overwrite the `nextSlide` slide passed in (for example:  `nextSlide = currentSlide`). The closest I got was using `slickGoTo` inside a `setPosition` callback but then I realised this was causing an infinite loop and would require me to manage some slick state stuff outside of slick.

adam-lynch:
Wait, it would all be so simple if I could just change `slidesToScroll` to `0`. I've been playing around with the code. I think this would work fine except you do a lot of things like `x / _.options.slidesToScroll` or `x % _.options.slidesToScroll` and I'm not 100% sure what to do in each case.

adam-lynch:
Ok, I've found a solution. If #1139 is merged, then I can use that along with disabling `swipe` temporarily. For example:

```javascript
var $form = $('form');
  var allow = false;

  $form.on('beforeChange', function(e, slick, currentSlide){ 
    if(!allow){
      e.preventDefault(); 
    }
  });   

  $form.slick({
      slide = 'fieldset',
      dots: true,
      infinite: false,
      speed: 300,
      slidesToShow: 1,
      arrows: false,
      swipe: allow
  });
  
  // Just so it doesn't look like two dots are active
  $form.on('focus', '.slick-dots button', function(){
    $(this).trigger('blur');
  });
  
  $('#allow').on('click', function(e){
    allow = !allow;
    $form.slick('slickSetOption', 'swipe', allow, true);
  });
```

So with this, initially you cannot change slide (with click or swipe) and you can once `#allow` is clicked.

simeydotme:
think about this logically....

Why on earth would you want to prevent the slideChange when you've already triggered the slideChange? - it doesn't make sense.

If you want to prevent changing slides, then run the logic before you change slides and change the carousel options to disallow interaction... do not allow user to change slides and then stop half way through.

adam-lynch:
>If you want to prevent changing slides, then run the logic before you change slides and change the carousel options to disallow interaction

But I didn't see any way to disable / pause slick. The closest thing is `unslick` but that removes the dots and styling. I don't want to have to clone the dots manually or anything while it's "disabled".

adam-lynch:
Ping

nfour:
[...]

PR #2104, opened 2016-02-07
Prevent slide change in beforeChange event using preventDefault

cfoehrdes:
This pull request adds the ability to prevent a slide change using `event.preventDefault()` related to the closed pull request #1139 and tries to solve #1138.

I moved the synchronization of `asNavFor` sliders right after the before change event to also prevent the change of linked sliders.

Demo Fiddle: http://jsfiddle.net/cfoehrdes/2sc7fuo9/3/

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -81,8 +81,8 @@ arrows | boolean | true | Enable Next/Prev arrows
 appendArrows | string | $(element) | Change where the navigation arrows are attached (Selector, htmlString, Array, Element, jQuery object)
 appendDots | string | $(element) | Change where the navigation dots are attached (Selector, htmlString, Array, Element, jQuery object)
 mobileFirst | boolean | false | Responsive settings use mobile first calculation
-prevArrow | string (html|jQuery selector) | object (DOM node|jQuery object) | <button type="button" class="slick-prev">Previous</button> | Allows you to select a node or customize the HTML for the "Previous" arrow.
-nextArrow | string (html|jQuery selector) | object (DOM node|jQuery object) | <button type="button" class="slick-next">Next</button> | Allows you to select a node or customize the HTML for the "Next" arrow.
+prevArrow | string (html \| jQuery selector) \| object (DOM node \| jQuery object) | `<button type="button" class="slick-prev">Previous</button>` | Allows you to select a node or customize the HTML for the "Previous" arrow.
+nextArrow | string (html \| jQuery selector) \| object (DOM node \| jQuery object) | `<button type="button" class="slick-next">Next</button>` | Allows you to select a node or customize the HTML for the "Next" arrow.
 infinite | boolean | true | Infinite looping
 initialSlide | integer | 0 | Slide to start on
 lazyLoad | string | 'ondemand' | Accepts 'ondemand' or 'progressive' for lazy load technique. 'ondemand' will load the image as soon as you slide to it, 'progressive' loads one image after the other when the page loads.
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2313,10 +2313,6 @@
             return;
         }
 
-        if (sync === false) {
-            _.asNavFor(index);
-        }
-
         targetSlide = index;
         targetLeft = _.getLeft(targetSlide);
         slideLeft = _.getLeft(_.currentSlide);
@@ -2371,7 +2367,17 @@
 
         _.animating = true;
 
-        _.$slider.trigger('beforeChange', [_, _.currentSlide, animSlide]);
+        var beforeChangeEvent = jQuery.Event('beforeChange');
+        _.$slider.trigger(beforeChangeEvent, [_, _.currentSlide, animSlide]);
+
+        if(beforeChangeEvent.isDefaultPrevented()) {
+            _.animating = false;
+            return;
+        }
+
+        if (sync === false) {
+            _.asNavFor(index);
+        }
 
         oldSlide = _.currentSlide;
         _.currentSlide = animSlide;
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1487,7 +1487,7 @@
             }
         } else {
             rangeStart = _.options.infinite ? _.options.slidesToShow + _.currentSlide : _.currentSlide;
-            rangeEnd = rangeStart + _.options.slidesToShow;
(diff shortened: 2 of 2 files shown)


simeydotme:
@cfoehrdes .. same thing happens that I experienced when I tried; the slider gets stuck mid-way between drags. (try dragging/swiping)

cfoehrdes:
Ok, I see the problem. I think we need to try to implement some sort of bouncing back or rewind of the animation in case of a prevented transition.

simeydotme:
also the "nextSlide" is incrementing each time due to Autoplay

simeydotme:
also this is broken:
http://jsfiddle.net/2sc7fuo9/4/

simeydotme:
```
    if( beforeChangeEvent.isDefaultPrevented() ) {
        _.animating = false;
        if (dontAnimate !== true) {
            _.animateSlide(slideLeft, function() {
                _.postSlide(_.currentSlide);
            });
        } else {
            _.postSlide(_.currentSlide);
        }
        return;
    }
```

If you can implement that, and then create a few edge-case test fiddles, with responsive options.
I will test and see if we can get this in before next release :)

cfoehrdes:
Thanks, I will add it and do some more testing!

cfoehrdes:
I found some issues with the current implementation. The `_.animateSlide` call you added in the approach above causes the slider to animate to the next slide even if the transition was prevented. Will try to animate back to the source slide.

simeydotme:
@cfoehrdes ... my code was fine...
https://github.com/cfoehrdes/slick/commit/fa3d58305cfe20ddfa7995bc9563bb4b481dfb24#diff-eaaa72f2e5f1d4974e4f1ae902717908L2373
You've made an error on this line... it should be: `isDefaultPrevented` .. not `defaultPrevented`

:smile:

cfoehrdes:
Sorry I was confused because my IDE was showing a deprecation warning for `isDefaultPrevented`

simeydotme:
don't let your software write your code ;P
any update on tests with other features / settings? :grin:

cfoehrdes:
We tested it a lot in context of our project and works like a charm. Will extend the fiddle with some additional cases today.

cfoehrdes:
Added some additional cases and tested after rebase. Everything works fine except the `rtl: true`. Any idea why the slider is not working at all after changing the direction?

simeydotme:
sorry @cfoehrdes , i didnt get enough time this weekend :)
I'll do my best to get to this sometime this week.

sascha-hendel:
I have a small feature request related to this issue: If Change event is stopped to do first some animation and than the Change is inititiated on animation complete callback, there is one problem:
beforeChange event has parameters currentSlide and nextSlide, but not the boolean parameter if Change is animated. So, if the Change will be finally initiated in the animation complete callback (with slickGoTo) this parameter is missing ...

gabskoro:
Hi, any news on this?

leggomuhgreggo:
Hey @cfoehrdes the reason rtl wasn't working is that it requires the `dir` attribute on the slider. [Updated the fiddle here](http://jsfiddle.net/uvtyebz7/).

I know there's a sizable release coming soon. If you wanted to try to get this merged in there, the first step would be to resolve the merge conflicts.

lookapanda:
Is there a way to trigger the animation after everything?
Like, I preventDefault it first, do my stuff, and then continue using slick.slickNext(); it goes to the next slide, but it doesn't do any animation

domanchi:
I found a very temporary workaround for people wanting to achieve this effect, before it's implemented the "right" way, involving exceptions to circumvent lack of event handling.

### 1. Throw an exception, when you want to prevent a slide.
```
$('.your-class').on('beforeChange', function(event, slick, currentSlide, nextSlide) {
    throw "some-custom-exception-here";
}
```
This will cause the exception to propagate through the libraries, preventing the transition to the next slide.

### 2. Catch the exception
```
window.onerror = function(msg, url, line, col, error) {
    if (error === "some-custom-exception-here") {
        return true;  //  silent the error, and keep functioning as normal
    }
}
```
This is the only place we can "catch" the exception, while still maintaining functionality of our website.

### 3. Minor change in the code base.
```
_.animating = true;
_.$slider.trigger('beforeChange', [_, _.currentSlide, animSlide]);
```
We need to flip the order of these lines (found in Slick.prototype.slideHandler), if not the exception will have the slick internal state constantly stuck on an animating mode, when that's not true. This way, the animating flag will only be true when the slide is about to change, and nothing else can stop it.

As I said, this is kind of a janky work-around, so this minor change is not worth a PR. But if you're wanting to emulate this feature of preventing transitions, hopefully this method will work for you.

mrigo:
Hi, any news about this update?

kristos80:
If you don't want just to prevent the sliding to specific slides, but to also go to another slide @domanchi 's code can be extended as:

```
$('.your-class').on('beforeChange', function(event, slick, currentSlide, nextSlide) {
    var theSlideToGo = nextSlide + 1; // Change according to your needs
    try{
      throw "some-custom-exception-here";
    }catch(e){
       $('.your-class').slick('slickGoTo', theSlideToGo);
       throw e;
   }
}
```

pakiam:
@cfoehrdes after prevented beforeChange, also afterChange event fires

sipy:
Hi, any updates on this? I am using slick 1.8.1 and e.preventDefault on beforeChange event isn't working...

kriscoleman:
What's the status on this??? anyone care to fix merge conflicts and get this merged in finally?

dedacpo:
I only want to reminder to people that are begginers, like me, that the CDN jsdelivr doesn't work for this change (event.preventDefault()). In order to use this functionallity it's necessary to use this cdn:

<script src="https://rawgit.com/cfoehrdes/slick/master/slick/slick.js"></script>

I took a few hours to realize that ^^"

cfoehrdes:
Closing this one, due to inactivity

sascha-hendel:
Why is "inactivity" a reason for closing an issue???

Christoph Föhrdes <notifications@github.com> schrieb am Mi., 13. Nov. 2019,
10:36:

cfoehrdes:
It is not an issue it is a pull request. As it is opened by me and nobody is working on it since 2 years I am closing it. If anyone is interested to work on this, feel free to open a new PR

Issue #2146, opened 2016-02-22
Something wrong with arrows in vertical mode


inferusvv:
This bug is reproducable in vertical mode with `infinite: false`

1. Open jsFiddle below.
2. Click on "Next" many times (for example 3 times)
3. Click "Prev".

"Prev" is clickable after N times you clicked "Next"

====================================================================

use this jsfiddle to reproduce your bug: 
https://jsfiddle.net/cx5jLnae/

leggomuhgreggo:
[Confirmed with the master](https://jsfiddle.net/4djkfavz/)

mazzaker:
Same behavior when not in vertical mode...

simeydotme:
its not "n times" it's a total of 3 times in @leggomuhgreggo fiddle, even if I press next 40+ times, still 3rd time on prev. :)

inferusvv:
@simeydotme Yes! You are write :)

simeydotme:
(edited previous comment, after trying with `slidesToShow: 5`

johannesdold:
Any updates here? We have several clients that are waiting for a patch for this bug. An update would be great. Any way we can help with that?

simeydotme:
closed by #2152 , please take the master branch, or wait for @kenwheeler to release the next version.

shivika:
+1

Issue #2035, opened 2016-01-16
Dots count not being correct


kaptenw:
Hi,
First of all, amazing plugin! This is not a high priority thing, but the dot count sometimes get's a bit weird. Below is an example where we have 16 slides, `slidesToShow` is set to `4` and `slidesToScroll` is set to `3` (don't even ask me why I have that setup). In the example an "extra" dot is added at the end that's not needed.

http://jsfiddle.net/6p10mvo8/1/

kenwheeler:
Thanks for playing by the rules, looking into this today

taylorKonigsmark:
It looks like it has to do with the getDotCount function, mind if I take a look into it today and submit a PR? I just started using github, and would love to contribute to this project.

kaptenw:
Thanks, Ken, I appreciate it!

Here's another example with `3` slides. Settings:  `slidesToShow: 2` and `slidesToScroll: 1`.
http://jsfiddle.net/6p10mvo8/2/

mazzaker:
Here the Fiddles with my fix
[first Fiddle](http://jsfiddle.net/6p10mvo8/13/)
[second Fiddle](http://jsfiddle.net/6p10mvo8/12/)

truefalse10:
having the same issue here:

[Fiddle with wrong dotCount](http://jsfiddle.net/H66vv/28/)

mazzaker:
Please change the slick.js with current master. Same answer on your last post.

PR #2152, opened 2016-02-23
Something wrong with arrows in vertical mode

mazzaker:
Fix for issue #2146 

[Fiddle](https://jsfiddle.net/7zq6m78d/2/)

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1049,7 +1049,9 @@
             }
         } else if (_.options.centerMode === true) {
             pagerQty = _.slideCount;
-        } else {
+        } else if(!_.options.asNavFor) {
+            pagerQty = 1 + Math.ceil((_.slideCount - _.options.slidesToShow) / _.options.slidesToScroll);
+        }else {
             while (breakPoint < _.slideCount) {
                 ++pagerQty;
                 breakPoint = counter + _.options.slidesToScroll;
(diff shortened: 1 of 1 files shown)


simeydotme:
thanks, mazzaker, could you do one more fiddle with a couple of "asNavFor" examples? :) ... the slideHandler method runs TWICE for asNavFor.

mazzaker:
hmm I'll push a new version soon ;)

the nav slider stops if last item is active. Thanks for the hint.

mazzaker:
Hmm there was something wrong with the math of getDotCount. The new version seems to fix the problem but there are so many ways to use slick this is really awesome ;)

This have to be tested with every settings available. Is there a complete test fiddle or something similar?
 
[Fiddle](https://jsfiddle.net/z6njw5z7/2/)

leggomuhgreggo:
@mazzaker here's one from another PR that has a bunch of the settings--just needs to be updated to use this branch: http://jsfiddle.net/gy1jy44j/2/

mazzaker:
[Fiddle with this branch](http://jsfiddle.net/gy1jy44j/4/)

There seems to be a problem with center mode and the number of items <= slidesToShow but this issue is the same on the master.

[Fiddle with master](http://jsfiddle.net/gy1jy44j/5/)

mazzaker:
One more #2035

simeydotme:
## ISOLATED TESTS

@mazzaker -> http://jsfiddle.net/a4ytf12b/
7a79c43f2864ba4087e2961849c2da7bdcfe2de8 **(last commit)** -> http://jsfiddle.net/5y7La0es/

I think it's good to go?

ayanapowell:
Im currently having the same issue. When i set `swipeToSlide: true`, i get too many dots. When set to `swipeToSlide: false`, i get the correct amount of dots...

Issue #2110, opened 2016-02-09
Images not lazy loading when infinite and previous button clicked.


srussking:
Hi, 

I've run across a strange bug that depends on a number of factors.

- Number of slides in the carousel (7 seems to show the break most clearly)
- Infinite scroll
- Lazy loading

If the stars align, the carousel will not lazy load the first two images after you hit previous. If you hit it again, they will load in and everything will be fine. But the first time you hit previous will get:
![screen shot 2016-02-09 at 10 33 10 am](https://cloud.githubusercontent.com/assets/182509/12926511/99b1c0c0-cf18-11e5-8e7a-fedceadc4d97.png)

We are using a placeholder image to keep the height stable as the new slides are loaded in, which is why the fiddle has a src and data-lazy, but this doesn't seem to be a factor on whether or not it attempts to load those images.

Fiddle
http://jsfiddle.net/srussking/8v9ubrcr/

simeydotme:
the stars wont align if you don't provide a functioning test-case :confused: :stars:

srussking:
Ugh, I have no idea how to save in their new system... maybe this one, seems to work when I try it on a different browser.
http://jsfiddle.net/srussking/8v9ubrcr/2/

unzsnu:
Same issue for me.
`$(".crsl").slick({infinite:true,lazyLoad:'ondemand',adaptiveHeight:true, touchMove: true, fade: true, cssEase: 'linear'});`
Going  back starting from the first image, it lazyloads the last image, but with height:0.
Setting `infinite: false` is a solution, but not for this problem.

simeydotme:
I think, at line `1496` the code needs to change to:

```js
if( _.currentSlide >= _.slideCount - _.options.slidesToShow ) {
    loadRange = 
        _.$slider.find('.slick-slide:not(.slick-cloned)')
            .slice(_.slideCount - _.options.slidesToShow, _.slideCount);
} else {
    loadRange = 
        _.$slider.find('.slick-slide')
            .slice(rangeStart, rangeEnd);
}
```

But I haven't done extensive testing on it.
I wonder if @leggomuhgreggo wants to try my code out and do a bunch of testing? ... lol ... I imagine this could possibly have regression impact on centerMode and variableWidth... :joy: :kiss:

leggomuhgreggo:
@simeydotme was thinking the same thing!

simeydotme:
[...]

PR #2164, opened 2016-02-25
Fixes lazy loading issue when previous button is clicked

leggomuhgreggo:
This PR is in reference to #2110 When previous button was pressed lazy loading didn't trigger.

[Here's a fiddle to demo](http://jsfiddle.net/leggomuhgreggo/fmo50w7n/252/). There are some issues with centerMode, but that's outside the scope.

### DIFF
--- a/slick.jquery.json
+++ b/slick.jquery.json
@@ -9,7 +9,7 @@
         "touch",
         "mobile"
     ],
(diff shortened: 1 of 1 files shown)


simeydotme:
:(

PR #2169, opened 2016-02-26
Add option for more fingers to be useable for swipe gesture

jonnybot0:
There are some contexts where it would be helpful to treat a swipe with any number of fingers just like a single-finger swipe.

My particular use case is a digital signage application. Users won't be using the normal multi-touch gestures that one might reasonably expect in a browser (two-finger swipe for "BACK", for example). As such, it can be helpful to allow more than one finger to perform the swipe action.

This fix still assumes that the 0th finger in the touch events is the one to track for purposes of the swipe; it's just a way to let someone swipe with two fingers by "mistake", which is particularly helpful for large touch-enabled surfaces.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -80,6 +80,7 @@ fade | boolean | false | Enables fade
 arrows | boolean | true | Enable Next/Prev arrows
 appendArrows | string | $(element) | Change where the navigation arrows are attached (Selector, htmlString, Array, Element, jQuery object)
 appendDots | string | $(element) | Change where the navigation dots are attached (Selector, htmlString, Array, Element, jQuery object)
+maxSlideyFingers | number | 1 | Change how many fingers can be used for a swipe gesture
 mobileFirst | boolean | false | Responsive settings use mobile first calculation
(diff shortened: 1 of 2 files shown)


simeydotme:
Appreciate the effort, @jonnybot0 , but you should have asked a question first really.
It was a deliberate decision for single-finger touches. Your use-case is like 0.0001% of the touch-enabled world. That's what Fork's are for :) -- or you can overload the touch method.

Best, Simon.

jonnybot0:
No prob, @simeydotme. I figured this was a niche case, but on the off chance that you would want to support it, I thought I'd put it out there, as I'd already written the code.

I probably will look at re-implement this with an overload, as if prefer to keep the slick library in my dependency list rather than maintaining a fork.

Anyway, thanks for giving it a look. I do like Slick!

AaronBuxbaum:
For what it's worth, there's a really good use-case for two-finger swipe: scrolling the carousel on desktops.

Try doing a two-finger horizontal swipe on Netflix, for an example.

PR #2199, opened 2016-03-16
Fixed an iOS Bug considering width

activenode:
On iOS resize Events are triggered when scrolling the page. This means that the slick-slider will get into that part of code where it checks if the old window width (_.windowWidth) has delta with the new one. This also means that the first time scroll happens on iOS the resize will be triggered. 

Furthermore: When you open a scrollable layer and hide all the elements below, the surrounding container of the slick-slider will have no width and height but will try to recaculate based on the fact scrolling happened. This results in unwanted slider results.

If you do _not_ accept this fix or if you need _.windowWidth = 0 for any other reason please tell me and I can provide a different fix for this problem.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -138,7 +138,7 @@
             _.transformType = null;
             _.transitionType = null;
             _.visibilityChange = 'visibilitychange';
-            _.windowWidth = 0;
+            _.windowWidth = $(window).width();
             _.windowTimer = null;
 
             dataSettings = $(element).data('slick') || {};


activenode:
Additional Information:
If and only if (!) this fix is not accepted for any further reasons a ":visbility"-based solution must be provided to avoid this bug.

simeydotme:
This looks like a sensible fix, but I cannot see the problem;
could you please provide a JS fiddle of before / after to demonstrate this problem? there's a fiddle to clone in the contribution guidelines, and you can replace the slick.js with your own version for the "after" using rawgit.com :)

activenode:
As of the fact that this requires the full page to be the script itself I do not see any way of providing this through CodePen / JSFiddle as I did not find a way to run the script without being iframed.

Is it ok when I provide a link on Github?

simeydotme:
You can provide a Fiddle, and then supply this link:
http://jsfiddle.net/simeydotme/fmo50w7n/embedded/ (example)
to break out of the iFrame :)

PR #2217, opened 2016-03-23
Modify adaptive height method to work with multiple visible slides

mapolin:


SimonShirley:
You might want to follow the comments in this thread as I changed the same block of code over a year ago and it has only just been commented on:

https://github.com/kenwheeler/slick/pull/1151

simeydotme:
sorry @mapolin this is not up to scratch, check #1151 for my full comment :)

PR #2226, opened 2016-03-31
added verticalDown feature

nomasgrim:


nomasgrim:
added verticalDown option. default is set to false.

the allows carousel to scroll downward instead of upward. the only catch to this is you have to set the slidesToShow to all of your slides minus the slidesToScroll you want. In your styles, you can overflow:hidden a desired height as to not show ALL the slides at once.

Best I could do with the time at hand.

Issue #1890, opened 2015-11-13
Slide blinks at the end of infinite-rotation-transition


noBlubb:
The transition from the last slide to the first slide (and inverse) sometimes causes an image within a slide to blink for a moment. I created a jsfiddle with minimal configuration to show this issue:
http://jsfiddle.net/f0up44wt/2/ (4-slides version)

Try to slide between Slide 1 and Slide 4 (to the left). At the end of the transition the image will eventually blink.
I spent the last 30 minutes sliding cats around and every time i think the bug is gone i see a flickering...check out the fun fact below...

On Chrome 46.0.2490.86 (64-bit) (Mac OS X 10.11.1) it happens:
- never, when moving "within the normal order" [1, 2, 3, 4], 1->2, 2->1, 2->3, 3->2...
- sometimes, but only when the previous transition is finished and only when moving between the last and the first slide in infinite scroll mode [1, 2, 3, 4], 4->1 (moving right), 1->4 (moving left)

My guess it's because of the changing transition offset in the root (-slick-) element, but DOM break points didn't really help me to figure it out.

*Fun fact* increasing the number of cats in the fiddle actually seems to resolve this problem:
http://jsfiddle.net/f0up44wt/3/ (10-slides version)

Please +1 if you can reproduce this issue, i'm going crazy :cat2:

ahmadalfy:
+1 I can reproduce it. Someone here possted a fix earlier using CSS. Will try to get it

noBlubb:
#1845 references #1626, but there is no animation for * in my fiddle.
Turns out  it runs great on Safari so I'll focus on issues with Chrome. 
I came up with http://stackoverflow.com/questions/3461441/prevent-flicker-on-webkit-transition-of-webkit-transform and will try to adapt this solution.

noBlubb:
The given indication was right, see this fiddle:
http://jsfiddle.net/f0up44wt/4/
The new backface-visibility-property could cause performance issues so handle with care.
https://css-tricks.com/almanac/properties/b/backface-visibility/

#1874 could be related, too.

noBlubb:
Small update:
The new rule actually fixed the flickering but eventually caused the last slide to disappear from time to time. I made a new fiddle following http://stackoverflow.com/questions/13952480/why-does-applying-webkit-backface-visibility-hidden-fix-issues-with-negativ and this works quite nice for me: http://jsfiddle.net/f0up44wt/5/

So all I needed to fix this was 
```.slick-slide {
[...]

PR #2230, opened 2016-04-02
Prevent flicker during slide transition.

### LINKED ISSUE #1961: Slide flickers when going from last to first slide
Have been researching possible fixes for this and just cannot figure out what's wrong. 

The issue seems to only be in chrome. But it doesn't happen in Chrome Canary.

http://jsfiddle.net/de2yzzds/6/ (use arrow keys to navigate and I also need the html this way)

The slide flickers when it goes from the last slide to the first slide

I'm on a mac using Chrome Version 47.0.2526.80 (64-bit)

Saw this solution https://github.com/kenwheeler/slick/issues/1890 but it doesn't seem to work.

Can anyone else recreate the issue?

sliderblink

damianwajer:
Fixes #1961.


simeydotme:
could you please provide a before/after fiddle (use the contributing guidelines for help) so we can test on a few devices, @damianwajer :)

damianwajer:
@simeydotme I'm not sure how to reproduce this bug with 100% accuracy because it seems that it occurs randomly depending on operating system and browser version (I couldn't find any specific pattern). Can we use examples from existing issues?

I've prepared fixed version of cases from:
#1961 
Before: http://jsfiddle.net/damianwajer/85a7z1vc/
After: http://jsfiddle.net/damianwajer/85a7z1vc/1/

#1890: 
Before: http://jsfiddle.net/damianwajer/cp26anvr/
After: http://jsfiddle.net/damianwajer/cp26anvr/1/

davidrhoderick:
Please merge this, I'm getting the white flash issue referenced by #1890

DanielRuf:
This is still happening.

DanielRuf:
https://github.com/kenwheeler/slick/issues/1961

niklasp:
The fiddles are using slick 1.5.8. It is not working for me in neither 1.7 nor 1.8

DanielRuf:
> The fiddles are using slick 1.5.8. It is not working for me in neither 1.7 nor 1.8

What exactly does not work for you? This issue is about a specific bug.
There may be breaking changes in 1.7 and 1.8.

PR #2255, opened 2016-04-17
Fix on demand lazy load range with initialSlide and slidesToScroll

rphelan:
The start of the visible range (for lazy loading purposes) isn't always currentSlide because if the slick is scrolled to the end and the slide count isn't divisible by the slidesToScroll, the scroll will be limited so as not to scroll past the end.  This change accounts for that case.


flyingL123:
Thanks for this. Was just trying to fix the same problem.

PR #2302, opened 2016-05-04
Replaced protocol-relative URL with HTTPS protocol

drmonkeyninja:
It is no longer considered good practice to use protocol-relative URLs (http://www.paulirish.com/2010/the-protocol-relative-url/). So the CDN request should be made via HTTPs.


markvantilburg:
@ahmadalfy this one can be closed, the current readme is already up to date.

ahmadalfy:
Closing as it's already updated

PR #2307, opened 2016-05-09
Charset declaration in theme prevents from importing within a selector

javiercejudo:
The motivation is to be able to import the theme within a selector while working around https://github.com/sass/sass/issues/2023

``` scss
.my-class {
  @import 'path/to/slick-theme-content';
}
```

The diff looks big but I just pulled the content onto its own file while leaving the original `slick-theme.scss` so that the CSS output is still the same.


rudzboy:
+1 !!!!!!!!!!!!!!

rudzboy:
Compiling with gulp :
`Error: It's not clear which file to import for '@import "slick"'.
       Candidates:
         slick.scss
         slick.css`

I really wish we can import the lib with selectors the way @javiercejudo suggested ;)
Please merge this ;)

Issue #2212, opened 2016-03-21
How to check was slide triggered automatically?


JILeXanDR1:
I'm using `afterChange `event for make some login in my application... Is there property like `isTriggeredAutomatically`?
For example:

```
            $('.slider').on('afterChange', function(event, slick, currentSlide) {
                if (slick.isTriggeredAutomatically) {
                    // if slider was run by timer, we don't do anything
                } else {
                    // do something
                }
            });
```

leggomuhgreggo:
Hmm not without some serious hacking I don't think. It's an interesting idea.

Closing: support request

JILeXanDR1:
@leggomuhgreggo 
I've added pull request https://github.com/kenwheeler/slick/pull/2327 to support this

leggomuhgreggo:
@JILeXanDR1 cool, I'll check it out now

PR #2327, opened 2016-05-20
Added  property "slick.isTriggeredAutomatically" in event handlers to…

JILeXanDR1:
Hi! This PR refers to my issue https://github.com/kenwheeler/slick/issues/2212

Usage is the same as I described above

```
$('.my-slider').slick({
    autoplay: true,
    autoplaySpeed: 1000,
    dots: true
}).on('afterChange', function (event, slick, currentSlide) {
    if (!slick.isTriggeredAutomatically) {
        // slider was run by user, do something!
    }
});
```

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -140,6 +140,7 @@
             _.visibilityChange = 'visibilitychange';
             _.windowWidth = 0;
             _.windowTimer = null;
+            _.isTriggeredAutomatically = false;
 
             dataSettings = $(element).data('slick') || {};
 
@@ -395,6 +396,8 @@
 
         var _ = this;
 
+        _.isTriggeredAutomatically = false;
+
         if (_.autoPlayTimer) {
             clearInterval(_.autoPlayTimer);
         }
@@ -405,6 +408,8 @@
 
         var _ = this,
             slideTo = _.currentSlide + _.options.slidesToScroll;
+
+        _.isTriggeredAutomatically = true;
 
         if ( !_.paused && !_.interrupted && !_.focussed ) {
 


leggomuhgreggo:
Firstly, thanks for the PR; it looks very straightforward and sensible. Detecting autoplay changes seems to be a recurrent use case and it's not easy to do as is.

Typically with PRs [we ask that you include some test cases](https://github.com/kenwheeler/slick/blob/master/CONTRIBUTING.markdown), but in this is low impact enough that it's more a formality. [Whipped one up with your fork just the same](http://codepen.io/leggomuhgreggo/pen/mPbeee).

I was apprehensive that this would only work in conjunction with the `beforeChange` event and not `afterChange`, thinking that `autoPlayClear()` would run before `afterChange` got triggered, but turns out it's subsequent; so all good there. 

Basically, it seems good to me, but @simeydotme or @kenwheeler will be the final verdict.



**The following is mostly just for reference:**
I think this satisfies all the use cases I've heard relating to trigger source identification, but in other issues on the subject people have suggested the possibility of specifically identifying the source. In assembling a list of things that cause slide changes, it occurs to me that the list is both sprawling and the implementation would complicate things a good deal; so I don't actually think it needs to be piggybacked here, but I'll include it at the end for reference purposes.

If this were to be done, my thinking is the new key in the global slick object would be called something like `slideChangedBy` instead of `isTriggeredAutomatically`, maybe default to `init` or something. Have to be careful about properly sequencing in relation to events. Possibly a bit messy.

**UI Events:**
1. autoplay
2. swipe
3. drag
4. arrow-left
5. arrow-right
6. slide-select
7. dot-select

**Public Methods:** 
1. slickNext()
2. slickPrev()
3. slickGoTo()
4. slickAdd()
5. slickRemove()
6. slickFliter()
7. slickUnfliter()

Was thinking about including `setOptions()` but it struck me as a crazy thing to try to fuss with the overlap it has.

PR #2331, opened 2016-05-24
fixed a swipe action bug

CaiWenlie:
When you swipe left up (or right down) fast, the swipe direction will be
"vertical", then the slick will do nothing, and the slick list will just
stop in the middle -- that's horrible! Fortunately this bug is fixed
now/


CaiWenlie:
slick is the best!

Issue #2020, opened 2016-01-13
why use aria-describedby in slides?


plweil:
I'm wondering about the appropriateness or need to use aria-describedby in each slide to refer to the navigation dots. The dots don't provide any sort of descriptive information about the slide.

I'm not which which aria attribute would be appropriate to indicate the relationship between the dots and the slide (if this is the goal). Possibly aria-controls? But aria-describedby doesn't seem right.

Moreover, this attribute is still generated even when the dots are not present , which is the default for Slick. So, unless I'm missing something, it shouldn't even appear by default.

ccbayer:
@plweil i agree; i had to fake this to pass validation by visually hiding the dots even though i really wanted them gone together. if the dots don't exist, then the describedby ID doesn't exist, which is invalid. 

additionally, i noticed this 'hack' didn't work on a "center" slick slider.
i'm guessing the cloned slides iterate their referred by ID - which it shouldn't, since they are clones.

jayburling:
I am also dealing with the validation errors caused by this.

wemaycode:
I came up against this issue as well. In my slideshow each slide has a div corresponding to the title of the slide. I set that div to the "aria-describedby" attribute generated by slick to get around the ADA compliance.
```javascript
$('.slide-title').each(function () {
    var $slide = $(this).parent();    
    if ($slide.attr('aria-describedby') != undefined) { // ignore extra/cloned slides
        $(this).attr('id', $slide.attr('aria-describedby'));
    }
});
```

ccbayer:
@wemaycode that's a good alternative / fallback!

surfbird0713:
I'm also having this issue, would be nice to have a setting for this or remove aria-describedby when dots are set to false.

benvoynick:
Also noticed this going through the markup generated. aria-describedby really shouldn't be set if the element it points to doesn't exist.

plweil:
And even if the dots exist, why use aria-describedby? The dots don't provide a description. If there is something we want to communicate to the user regarding the relationship between the slide and the dot, let's use the aria attribute that accomplishes what we want.

leggomuhgreggo:
Good question! I'm not an expert, and I don't have strong feelings, but my guess is that the describedby is less in reference to the dot representation, and more in reference to the markup, in which there's a number corresponding to the slide it navs to. The number is hidden with the theme styles and replaced with that dot. Moreover for implementations that use custom paging, there's often even richer relationship information conveyed.

[...]

PR #2346, opened 2016-05-31
Aria describedby Tags - Hide if dots are disabled, evenly distribute  when more slides than dots

ibarsi:
## Description

I have identified 2 scenarios in which I believe there are issues in the way "aria-describedby" tags are generated and applied to slides.

1. When the "dots" setting is turned off, "aria-describedby" tags are still added to slides, but are referencing the "would be" ids that the dots DOM elements would have if they were enabled. This breaks WAVE and A Checker tests, as those tags must reference the ids of elements that exist in the DOM.

2. When there are more slides than there are dots, "aria-describedby" tags are still generated and applied to slides as if there were dots available for each slide (ex. if there are 2 dots but 6 slides, each slide has an "aria-describedby" tag with a unique id, 4 of which do not actually exist as dots on the page). This also breaks WAVE and A Checker tests for the same reasons.

**Issue:** #2020 

## Resolution
To resolve these issues, I've applied the following changes:

1. When "dots" is set to false, I do not add "aria-describedby" tags to the slides, as there are no longer any elements that could reasonably describe the slides (see the note below for my opinion on this).

2. When there are more slides than there are dots, I set every visible slide to the currently selected dot. This should ensure that all slides have an "aria-describedby" tag that points to an existing DOM element.

**Fiddle:** https://jsfiddle.net/ibarsi/4khqcw9t/5/

### NOTE
I just wanted to mention that we may want to consider removing the generation of "aria-describedby" tags entirely... After these changes, accessibility checkers won't give the DOM generated by slick a failing grade, however I don't believe the dots provide a useful description of their referenced slides. Elements referenced as describers should really contain some sort of text that provides screen readers additional details that may not be apparent. This should probably be something that users of slick tag and create themselves, if they choose to.

**Spec Reference:** https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1286,10 +1286,11 @@
         _.$slideTrack.attr('role', 'listbox');
 
         _.$slides.not(_.$slideTrack.find('.slick-cloned')).each(function(i) {
-            $(this).attr({
-                'role': 'option',
-                'aria-describedby': 'slick-slide' + _.instanceUid + i + ''
-            });
+            $(this).attr('role', 'option');
+            
+            if (_.options.dots === true) {
+                $(this).attr('aria-describedby', 'slick-slide' + _.instanceUid + i + '');
+            }
         });
 
         if (_.$dots !== null) {
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1288,8 +1288,11 @@
         _.$slides.not(_.$slideTrack.find('.slick-cloned')).each(function(i) {
             $(this).attr('role', 'option');
             
+            //Evenly distribute aria-describedby tags through available dots.
+            var describedBySlideId = _.options.centerMode ? i : Math.floor(i / _.options.slidesToShow);
+            
             if (_.options.dots === true) {
-                $(this).attr('aria-describedby', 'slick-slide' + _.instanceUid + i + '');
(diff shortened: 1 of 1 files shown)


dgpokl:
My vote: :+1: Strongly agree with the "note" above as well. Thank you @ibarsi - I'm going to use your fork for now; hoping to see it merged in!

leggomuhgreggo:
Looks good to me. Thanks @ibarsi. Appreciate the assessment in your note.

[Test for good measure](http://codepen.io/leggomuhgreggo/pen/xORggE)

ibarsi:
No problem, glad my contribution helped!

sassycoder:
Hey, many thanks for this. I had to add another condition in case the dots aren't rendered on initial page load e.g. dots not needed at desktop, but will be needed in mobile:

`if (_.options.dots === true && _.slideCount > _.options.slidesToShow) {`

signal-intrusion:
How close are we to having this included in a release?

tyleryoungblood:
> How close are we to having this included in a release?

Looks like [2346](https://github.com/kenwheeler/slick/pull/2346) was included in the [1.71](https://github.com/kenwheeler/slick/releases/tag/1.7.1) release on Jul 30, 2017.

PR #2355, opened 2016-06-07
add condition for aria-describedby

janein:
do not add aria-describedby when dots are disabled

should solve issue #2020


leggomuhgreggo:
Hey @janein thanks for the PR. This looks to be essentially the same as #2346. I'm going to close this one, but if I missed some compelling advantage of yours please don't hesitate to set me straight. Appreciate the contribution!

PR #2365, opened 2016-06-15
Slick dots and arrows don't show, Unicode problem.

alegalviz:
Master version. AngularJS, Grunt, Bower. 

In minify process dots and arrow don't show. @charset mut be with double quotes.


XhmikosR:
@kenwheeler this is a pretty important issue. The spec clearly states double quotes.

PR #2369, opened 2016-06-16
Option for using up and down arrow keys.

senn-lb:
Added the option _verticalKeys_ for when you want to change the arrow keys to up (prev slide) and down (next slide) when you have _vertical slide mode_ enabled.


ewalpole:
Semantically, I don't think naming them prev and next key actually make sense in the code if rtl is enabled (because then next is prev and vice versa if I am not mistaken), but that is just nitpicky. I was looking for someone who had fixed this and those changes are similar to the ones I made.

PR #2405, opened 2016-07-06
Fix tab behavior for accessibility

eirslett:
Tab selection was not possible when tabbing
through an element that was outsideo of the
current view.

This changes the behavior, so that you can
tab through all the slides if they have an
element that can be focused (e.g. <a> link).

When you tab to a slide outside of the current
view, slick will automatically switch its
position so that the slide comes into view.

I'll try to make a jsfiddle that reproduces the issue.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1278,9 +1278,6 @@
         var _ = this;
         _.$slides.add(_.$slideTrack.find('.slick-cloned')).attr({
             'aria-hidden': 'true',
-            'tabindex': '-1'
-        }).find('a, input, button, select').attr({
-            'tabindex': '-1'
         });
 
         _.$slideTrack.attr('role', 'listbox');
@@ -1393,6 +1390,7 @@
 
         if (_.options.accessibility === true) {
             _.$list.on('keydown.slick', _.keyHandler);
+            _.$list.on('focusin a, button, input, select', $.proxy(_.focusSlideHandler, _));
         }
 
         if (_.options.focusOnSelect === true) {
@@ -1449,6 +1447,15 @@
             }
         }
 
+    };
+
+    Slick.prototype.focusSlideHandler = function(event) {
+        event.preventDefault();
+        event.stopPropagation();
+        var _ = this,
+            slide = $(event.target).closest('.slick-slide'),
+            index = slide.data('slick-index');
+        _.slideHandler(index);
     };
 
     Slick.prototype.lazyLoad = function() {


goetsu:
Not a good idea I think because it Will force user to go trougth each tabs instead of letting him choose one. Imagine you force mouse user to click on every slide before to be able to read the rest of the page

eirslett:
No forcing. With screen readers, you can easily jump/skip, e.g. from one `<h1>` to the next `<h2>` to the next `<h2>` etc, and also jump between sections of the site like `<nav>` or `<article>`. Tabbing through every single element is only something you do if you really want to.

goetsu:
Not if you are simply using keyboard navigation and not a screenreader user

eirslett:
I would think "power users" of keyboard navigation usually have their keyboard shortcuts setup as they need them.
More importantly, the only thing that's more annoying than having to tab through elements, is not being able to interact with an element at all (without a mouse). E.g. in a carousel like "featured articles" (like the frontpage of a newspaper or magazine), you would need to be able to select every element, in order to navigate to the full article.

goetsu:
I'm not aware of any shortcut for keyboard only user to skip a block of content. Futhermore i don't say that slide must not be focusable but only that it must be the same as for mouse user. If you want to go next prev use the next prev button. If you want to go to specific slide use dots navigation. Then the focus must be moved on the current slide but only on that one.

eirslett:
> I'm not aware of any shortcut for keyboard only user to skip a block of content. 

This is how you do it with a screen reader: http://www.chromevox.com/keyboard_shortcuts.html

I agree that it should be "the same" as mouse over, in the sense that tab navigation is basically the same as mouse over (I think).

Carousels are fundamentally inaccessible unless you are sighted and can use a mouse to navigate; it's the combination of multiple pages, multiple slides per page, prex/next buttons and dots, they all make sense if you can see the visual layout in front of you, but not so much if you cannot. Then you get a lot of cognitive overhead, which is tiring. (As opposed to simply tabbing through the items one-by-one). Also, keyboard navigations users are already used to the habit of tabbing through lots of elements on every page; the usual helpful trick for that is for website developers to implement a "skip to main content" link.

> If you want to go to specific slide use dots navigation. Then the focus must be moved on the current slide but only on that one.

Yes, it's possible, but very annoying, because you don't know what's "hidden behind" each dot until you navigate into it, so you'll have to jump back and forth between slides and dots to iterate through the elements.

Which user group do you have in mind exactly? Sighted users that choose to navigate with the keyboard?

goetsu:
I know how to do it with screenreader I mean for keyboard only user.

> Then you get a lot of cognitive overhead, which is tiring. (As opposed to simply tabbing through the items one-by-one)

tabbing true lot of slides can be tiring and cognitive overload too

> because you don't know what's "hidden behind" each dot until you navigate into it

that's why you need to have explicit label for dots and prev / next button especially if you use infinite carousel (there another pull request for that)

> Which user group do you have in mind exactly? Sighted users that choose to navigate with the keyboard?

yes or low vision user or anyone, as I already say for me no reason to impose reading or going trough the whole carousel when mouse user don't have to. 
Another reason to do it that way is that there is others issue/pull request about using aria tabs design pattern for slick. If you look at aria design pattern for tabs user don't go trough the whole tabpanel simply by using tab key

PR #2418, opened 2016-07-12
Added "ariaPolite" options to enable/disable the aria-live attribute

fantarama:
On some page templates the aria-live="polite" attribute make the entire page not accesible with screen readers because it force the screen reader to read the slick content on changes and users could not focus on other page fragments.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -39,6 +39,7 @@
 
             _.defaults = {
                 accessibility: true,
+                ariaPolite: true,
                 adaptiveHeight: false,
                 appendArrows: $(element),
                 appendDots: $(element),
@@ -501,7 +502,7 @@
 
     Slick.prototype.buildOut = function() {
 
-        var _ = this;
+        var _ = this, wrapperTemplate;
 
         _.$slides =
             _.$slider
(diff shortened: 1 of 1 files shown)


cshold:
This is a decent quick fix, though I think the ideal solution is to add/remove `aria-live="polite"` when the slideshow or an element in it is focused. That way a screen reader will announce the content changes only when you're interacting with the element.

fantarama:
Good suggestion but I think is not the scope of my fix when add/remove the `aria-live`: the property only enable/disable the attribute.

goetsu:
I agee that removing aria live is a good idea because without a pause button it Will force screenreader user to listen to the update each time it occurre.
Adding / removing it dynamically when moving focus on slide or inside it is not usefull because if you move focus it Will already read the content without the aria live (or at least user Will start to read it by himself) so in case you have an extra aria live you may have the content read twice.

My opinion is that we must remove aria live or have an option to remove it (so this pull request is good for that) and move focus on the slide when using dots or prev next buttons (need a specific pull request for that part)

goetsu:
And when having more than one block of content per slide the focus must go on the first block of content of the new serie of blocks when

PR #2442, opened 2016-07-28
A11y helpers

cshold:
- Remove `outline: none` on arrows/dots
- Comment out a `stopImmediatePropagation` as it interferes with custom focus work ([see this comment](https://github.com/kenwheeler/slick/commit/785d1698d157e13067b800ff902ff495d954670c#commitcomment-18434941))
- Change dot text to `Slide x` from simply a number

### DIFF
--- a/slick/slick-theme.css
+++ b/slick/slick-theme.css
@@ -38,7 +38,6 @@
 
     color: transparent;
     border: none;
-    outline: none;
     background: transparent;
 }
 .slick-prev:hover,
@@ -47,7 +46,6 @@
 .slick-next:focus
 {
     color: transparent;
-    outline: none;
     background: transparent;
 }
 .slick-prev:hover:before,
@@ -162,13 +160,7 @@
 
     color: transparent;
     border: 0;
-    outline: none;
     background: transparent;
-}
-.slick-dots li button:hover,
-.slick-dots li button:focus
-{
-    outline: none;
 }
 .slick-dots li button:hover:before,
 .slick-dots li button:focus:before
--- a/slick/slick-theme.less
+++ b/slick/slick-theme.less
@@ -50,9 +50,7 @@
     transform: translate(0, -50%);
     padding: 0;
     border: none;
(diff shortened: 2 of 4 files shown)


cshold:
cc @Shopify/themes-fed

cshold:
cc @stevebosworth @m-ux

leggomuhgreggo:
The current thinking is to maintain the `outline: none`, owing to the outline's design obtrusiveness. Easy enough to comment out during implementation if it is a requirement.

@simeydotme do you remember what was going on with the `stopImmediatePropagation` stuff?

cshold:
I totally meant to open this PR to my forked version, sorry about that. That said, I think it's worth considering these changes for the main repo as well.

m-ux:
@cshold it might be worth explaining some of the reasoning behind these changes.  

As a bit of a background, we're developers @ Shopify, and we are using your plugin in our Themes (awesome work BTW) - however we've been having some issues with a11y support.   

Carson could likely speak to that in more detail.

m-ux:
also: re `outline: none`'s obtrusiveness http://www.outlinenone.com/

Obviously, totally at your discretion but from an accessibility standpoint, this is something that is being strongly advised against these days.

cshold:
Michael said it best, `outline: none` is more obtrusive than it is helpful. It's impossible to undo that style so having it in the core offering isn't ideal.

As for `stopImmediatePropagation` (https://github.com/kenwheeler/slick/pull/2442/files#diff-eaaa72f2e5f1d4974e4f1ae902717908R1014), I can't tell why that was added in the first place. It's enough to disrupt any external JS that tries to access elements inside the slideshow, such as:
```
$('.slideshow`).on('focusin', function(evt) {
  if ($('.slideshow').has(evt.target).length) {
    $('.slideshow').slick('slickPause');
  }
});
```

cshold:
I'm going to open a more flushed out a11y update PR shortly. Closing this one.

PR #2456, opened 2016-08-06
Added lazyLoad option 'anticipated' to preload {n} images

Rendez:
### Description
So that the user experience is better, the images should be loaded on demand, but anticipated, meaning, that the slider will always be one step ahead of the user, by pre-loading the 1 next and 1 previous image so that the loading is seamless, without ever making the user aware that the image was loading. On the other hand, this represents an improvement over loading all the images progressively, as the amount of data to load might not be worth causing also a poor experience, specially on mobile devices with slow connections.

### Demo

http://jsfiddle.net/zeL1kmo3/12/

### Process

There are a few edge cases in which lazyLoad === 'anticipated' might not work as expected, they are rooted on the nature of the current implementation, some bugs and limitations:

1. If slidesToScroll > 1, going from index 0 to -2 and index 8 to 10 only contains one `cloned` item, so scrolling 2 shows an empty image during the animation This can be see in line https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L2034, as `slidesToShow` is used where `slidesToScroll` should be the amount of clones to set-up.

2. If slidesToScroll > 1 and the number of images is an odd number. Similar problem to 1.

3. When `fade == true`, the rangeEnd seems to be +1 count too much, this is either related to another bug or an implementation feature I am not aware of.

Credits to @AlaaSarhan for his collaboration bringing this feature to life.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -64,7 +64,6 @@
                 infinite: true,
                 initialSlide: 0,
                 lazyLoad: 'ondemand',
-                preLoad: false,
                 mobileFirst: false,
                 pauseOnHover: true,
                 pauseOnFocus: true,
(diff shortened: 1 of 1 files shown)


tomhicz:
Thanks for this - Very useful. 
I'm using it on a site in development and it seems to be working well so far. (not using slidesToScroll > 1 or fade.)

rafalf:
@kenwheeler why is that not present in the CDN-ed version of slick.js - **Version: 1.6.0** 
e.g. https://cdnjs.cloudflare.com/ajax/libs/slick-carousel/1.6.0/slick.js

Rendez:
@rafalf I have no rights to merge this PR.

rafalf:
@Rendez Thanks! we added these bits and pieces to 1.6.0 and stopped using cdn but cloudfront instead and working like a charm.

leroy0211:
Please merge this as soon as possible!

kenwheeler:
thanks @Rendez

swinggraphics:
Can this feature be added to the documentation?

PR #2460, opened 2016-08-10
Accessibility improvements

cshold:
Improves the overall accessibility of the slideshow in a few ways.
### [Demo](http://codepen.io/cshold/full/PzXGxG)

Each feature is split up by commit:
### 1. Remove `outline: none` on arrows and buttons

Removing this appeases the visual state of the slider but [ignores accessibility](http://www.outlinenone.com/). @leggomuhgreggo mentions a fix is to remove from the production styles if it's an issue, but I'd argue this is the best slider on the market and should make the proper decisions out of the gate — including accessibility.
### 2. Update the labels on arrows and dots

Currently the arrows say "Next" and "Previous". I've updated to "Next slide" and "Previous slide" to each of of them so it's more obvious what is happening when the buttons is pressed. Similarly, the dots used to be labeled with a number corresponding to their index. The new labels will say "Slide 1", "Slide 2", etc.
### 3. Updated role and aria attributes

[aXe Chrome plugin](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd) helps debug accessibility issues and shows a few on the current implementation:
- ARIA attribute is not allowed: `aria-selected="true"` on dots
  - This is because dots are currently `role="presentation"`. Changing them to `tab` fixes this (see below)
- Required ARIA child role not present: tab
  - The dot container is `role="tablist"` so the children need to be `role="tab"` (not `presentation`)
- Invalid ARIA attribute value: aria-controls="navigation00"
  - Only an error because `aria-controls` needs to have a matching `id` on the slide container. Fixed by:
    - Change `navigation00` to `slickSlide00` for dots `aria-controls`
    - Add `id="slickSlide00"` to the slide container (not cloned versions)
    - Change `aria-describedby` to `aria-labelledby` as it is the recommended use for `role="tablist"`. [See this demo](http://accessibility.athena-ict.com/aria/examples/tabpanel2.shtml)
- No need for `role="listbox"` on the slide track now that `tablist` is used properly
- Remove JS adding `role="button"` on dots since it's already in `_.defaults.customPaging`
- Remove JS adding `role="toolbar"` to the closest div after the dot container as it doesn't actually do anything
- Remove `aria-hidden` from dots. They should always be accessible by keyboard/screen reader.

Lastly, slightly reorganized the `ADA` functions as well. Now `initADA` is only called once on `init`, then on `postSlide` use a new function called `updateADA`. This reduces the amount of `initADA` JS running after each slide transition.

@kenwheeler 
cc @stevebosworth @m-ux

### DIFF
--- a/slick/slick-theme.css
+++ b/slick/slick-theme.css
@@ -38,7 +38,6 @@
 
     color: transparent;
     border: none;
-    outline: none;
     background: transparent;
 }
 .slick-prev:hover,
@@ -47,7 +46,6 @@
 .slick-next:focus
 {
     color: transparent;
-    outline: none;
     background: transparent;
 }
 .slick-prev:hover:before,
@@ -162,13 +160,7 @@
 
     color: transparent;
     border: 0;
-    outline: none;
     background: transparent;
-}
-.slick-dots li button:hover,
-.slick-dots li button:focus
-{
-    outline: none;
 }
 .slick-dots li button:hover:before,
 .slick-dots li button:focus:before
--- a/slick/slick-theme.less
+++ b/slick/slick-theme.less
@@ -50,9 +50,7 @@
     transform: translate(0, -50%);
     padding: 0;
     border: none;
-    outline: none;
     &:hover, &:focus {
-        outline: none;
         background: transparent;
         color: transparent;
         &:before {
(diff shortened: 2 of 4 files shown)


goetsu:
@cshold 

there is still an issue with the aria-label on prev/next button .
For example if your carrousel have 6 slides when user are on the slide 6 the label "next slide" make user think that there is a slide 7. 
The best solution that I see is to use the same label as on dots "slide x" or add an tabindex="-1" and aria-disabled="true" on button "next" when user reach the last slide and on button "prev" when user reach the first slide to prevent infinite loop.

Regarding using tablist/tab on dots and tabpanel on slides. I'm not a fan of this solution but if you want to go that way you also need to remove the role="option" on slide and follow the keyboard design pattern for tabs (only the current dots is focusable by default and you need to use left/right arrow key the reach the prev/next dots and update the currently shown tabpanel example http://a11y.nicolas-hoffmann.net/carrousel/

PR #2494, opened 2016-08-25
Implementing classes using the new ES6 syntax

leonardo-silva:
Hi,

In my PhD research, I am using slick in some experiments.

As a result, I detected the existence of a main class in the current codebase, mostly to implement the methods of the API (activateADA, addSlide, animateHeight, animateSlide, ...).

So, I decided to rewrite this code to follow the new syntax for classes provided by ECMAScript 2016, which I think is much more readable.  
All changes are just syntactical. I was also very careful to preserve the code style.

Do you think it is worth to integrate these changes to the project?

Thank you

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -34,2872 +34,2874 @@
         var instanceUid = 0;
 
         class Slick {
-	        constructor(element, settings) {
-
-	            var _ = this, dataSettings;
-
-	            _.defaults = {
-	                accessibility: true,
-	                adaptiveHeight: false,
-	                appendArrows: $(element),
-	                appendDots: $(element),
-	                arrows: true,
-	                asNavFor: null,
-	                prevArrow: '<button type="button" data-role="none" class="slick-prev" aria-label="Previous" tabindex="0" role="button">Previous</button>',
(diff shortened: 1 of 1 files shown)


leggomuhgreggo:
@leonardo-silva thanks for the PR! You may be interested in the v2 branch, where @kenwheeler's  started swapping to ES6 syntax, with component modules, and using webpack as a build system. I think it will be pure JS too so no jQuery dependency. If you look the last update was a while ago, but word is v2 dev is about to gear up again.

I'm curious what kind of research you're using the library for?

kenwheeler:
Whoah this is bananas. I'll check this out locally later, the site won't let me view the diff

leonardo-silva:
@leggomuhgreggo Very interesting! I will take a look at the v2 branch. Thanks.

Generally speaking, my research is about reverse engineering techniques for JavaScript systems. I first found slick when I was looking for popular JS systems (sorted by number of stars) on GitHub.

leggomuhgreggo:
Oh that's rad. Did you RE the slick js?

leonardo-silva:
Yes I did. Applying RE I can for example identify the API methods of an app without having to look at the source code. I like this kind of thing. :-)

PR #2510, opened 2016-09-08
Include stylus

svarnoi420:


Leolik:
@kenwheeler please, add stylus support for slick
@svarnoi420 thanks

Issue #1252, opened 2015-04-19
Prevent Carousel Sliding while drag-scrolling vertically on a touch device


Defcon0:
Hello,

the title says it all: This should be possible. Is it a bug or already a feature?

Thanks in advance.

Bye Defcon0

j-kwon:
+1 I have the same issue.

simeydotme:
set option: `verticalSwiping` to `false` ... although I'm super confused, because it defaults to `false` ;)

Defcon0:
I think, you didn't quite get the issue ;-) When I *first* scroll down in order to get to the stuff at the bottom of the page, and *then* swipe to the left or the right, it's then possible to scroll *and* swipe to left and right at the same time, which is ugly.

simeydotme:
Nice Video! very clear.

levymetal:
I just started using Slick and I have the same issue. Apart from that, it's incredible!

levymetal:
I've patched this in my own fork, but I've only tested it in my own personal project. If I get time, I'll try and test out all variations and submit a PR. 

Anyway, in the meantime, for anyone who wants a quick fix you can view my changes here: https://github.com/levymetal/slick/commit/36fee6b5ddb51081658e1aa4195a249793f77ddf

Rrrapture:
+1

Defcon0:
@levymetal: Awesome job :+1: I just integrated your code and it works like a charm. Hopefully, kenwheeler adds this in no time.

simeydotme:
@levymetal is it possible for you to create a JSFiddle with your source for me to play with ? :)

adg29:
+1

ansman:
Any news on this, it's really annoying when using a mobile device.

levymetal:
@ansman Try using my fork here: https://github.com/levymetal/slick.

This works fine for me, but I've not had the time to create a JSFiddle with all possible variations, so I don't wish to submit a PR unless I know it won't break anything.

colorfool-me:
+1

hyeyoungkang:
:+1: 
Awesome!!
It's a difficult problem to me.. :(
Thank you!!

jhpmatos:
+1

floriancapelle:
+1

jansson:
+1

leonelcbraz:
+1

theexplay:
+1

marcelombc:
+1

[...]

PR #2521, opened 2016-09-21
[fix] https://github.com/kenwheeler/slick/issues/1252

marcelombc:
## Description

This PR fixes the problem of carousel sliding while drag-scrolling vertically on a touch device. 
When the user scroll down and then swipe to the left or the right, the carousel swipes to left and right while scrolling which is quite annoying.
It's possible to see a video of the problem [here](https://app.box.com/s/778ipwoe0i1t474hla2lmf44b3j88lvk) and you can find the github issue [here](https://github.com/kenwheeler/slick/issues/1252)
### Demo

[Without the fix](http://jsfiddle.net/cvymm987/3/)
[With the fix](http://jsfiddle.net/jfex0dbh/1/)
### Note

You need to test these demos on touch devices
## Credits

All the credits goes to @levymetal. Thank you for the code to fix this issue.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2568,7 +2568,7 @@
             _.$slider.trigger('edge', [_, _.swipeDirection() ]);
         }
 
-        if ( _.touchObject.swipeLength >= _.touchObject.minSwipe && (_.swipeDirection() == 'left' || _.swipeDirection() == 'right') ) {
+        if ( _.touchObject.swipeLength >= _.touchObject.minSwipe ) {
 
             direction = _.swipeDirection();
 
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -104,6 +104,7 @@
                 loadIndex: 0,
                 $nextArrow: null,
                 $prevArrow: null,
+                scrolling: false,
                 slideCount: null,
                 slideWidth: null,
                 $slideTrack: null,
@@ -111,6 +112,7 @@
                 sliding: false,
                 slideOffset: 0,
                 swipeLeft: null,
+                swiping: false,
                 $list: null,
                 touchObject: {},
                 transformsEnabled: false,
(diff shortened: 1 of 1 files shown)


simeydotme:
I haven't tried this, sorry I'm not able to do a full review... however, just looking at the commited changes it seems logical... but throwing a question out there:

> Would this kill vertical carousels ?

marcelombc:
That's a good point @simeydotme. I had a small issue with that but is already fixed and the vertical and horizontal carousel is working has expected. Sorry for not test that scenario.

semenets:
I had the same problem today. When this PR will be ready for merge? :)

bbodien:
Would also ❤️ to see this merged.

dangelion:
@kenwheeler please merge this. Thanks

kenwheeler:
@leggomuhgreggo can you verify this is all good?

leggomuhgreggo:
Happy to -- I'll holler at it this evening.

leggomuhgreggo:
Apologies for the long delay. This passes all my testing, nice work @marcelombc 

In answer to @simeydotme's vertical mode question, I checked and vertical mode doesn't currently support touch swipe (at least on my phone), and, in any case, vertical mode has only ever been partially supported; so we're good here I'd say.

Merging. Thanks everyone!

FullstackJack:
Both examples are janky on Android. It seems that the fixed version prevents side to side scroll as well if you swipe too frequently.

leggomuhgreggo:
@FullstackJack can you elaborate on what you're seeing? I think a slide transition has to finish, before a new one will register; so, I wonder if that's what's goin on.

librisius:
@FullstackJack it still doesn't work for Android!

yanwsh:
This fix make swap very difficult on mobile device, you have to make exact horizontal swap, otherwise it won't work, this is poor user experience. Hope it can provide an option let user setup vertical swap length. Right now it's fixed. 

```javascript
            if (!_.options.verticalSwiping && !_.swiping && verticalSwipeLength > 4) {
                _.scrolling = true;
                return false;
            }
```

rssteffey:
I have to second what @yanwsh said above.  This fix has made scrolling a fullscreen div incredibly frustrating. (On an iPhone X, at least).  Are there any plans to add a tolerance threshold to this limitation?

gstoyanov:
I can confirm the problem @yanwsh mentioned above. Found an open issue about it as well: https://github.com/kenwheeler/slick/issues/3464

PR #2524, opened 2016-09-22
Add semicolon at start

octavioamu:
A good practice when working with multiple JS files is to add a single semicolon at the very beginning of each file to terminate any preceding malformed logic.


danieltj27 (approved):
Probably a good idea considering this file will be concatenated into a minified js file with other scripts. No harm in doing this.

octavioamu:
Thanks buddy

PR #2539, opened 2016-09-27
Fix less comment syntax

soooooot:
single line comment `// xxx` is a valid (official valid) syntax.
better to change to normal `/* xxx */` comment syntax.

p.s.  processor like `postcss` doesn't support the single line comment originally.

### DIFF
--- a/slick/slick-theme.less
+++ b/slick/slick-theme.less
@@ -1,6 +1,6 @@
 @charset "UTF-8";
 
-// Default Variables
+/* Default Variables */
 
 @slick-font-path: "./fonts/";
 @slick-font-family: "slick";


danieltj27:
It's a valid syntax for less and shouldn't break anything. Pretty sure the LESS guidelines state the two forward slashes should be used over the 'multi line' comment for a single line.

soooooot:
@danieltj27 You're right, it's valid for less, I thought it was invalid for my `postcss` processor doesn't recognise it. Other comments in the file are written as multi line style, should we follow the style for compatibility?

markvantilburg:
@ahmadalfy can you merge this, aligns with the other css comments.

ahmadalfy:
Great catch; merged

Issue #2339, opened 2016-05-26
slick-carousel/slick.scss is imported twice in rails-assets


shobee:
Hi there,

when using `slick-carousel` via `rails-assets` the file
`rails-assets-slick-carousel-1.6.0/app/assets/stylesheets/slick-carousel.scss` looks like this:

```ruby
@import 'slick-carousel/slick.scss';
@import 'slick-carousel/slick.scss';
```

My first address for this problem was of course https://github.com/tenex/rails-assets and @hut8 fortunatly find really quick the cause. Please see https://github.com/tenex/rails-assets/issues/352#issuecomment-221734934 for the details.

PR #2556, opened 2016-10-10
Update bower.json to only declare one main file per filetype

wolfika:
The `bower.json` file specification now explicitly states, that the `main` field should only contain the main entry points of the package, one file per filetype. In case of Slick, not only a single file was specified as the main stylesheet file, but `slick/slick.css`, `slick/slick.less`, and `slick/slick.scss` as well. I think the SCSS version can be considered the best case scenario.

refs #2339

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,8 +2,6 @@
     "name": "slick-carousel",
     "main": [
         "slick/slick.js",
-        "slick/slick.css",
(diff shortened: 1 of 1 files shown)


shobee:
@kenwheeler Are there any reasons why this fix was never merged? Today I ran into this problem again -.-

Issue #2555, opened 2016-10-10
Deprecated version field in bower.json file


wolfika:
Slick's `bower.json` file uses the `version` field which is now deprecated. Bower now reads the versions from git tags. Related part of the `bower.json` [specification](https://github.com/bower/spec/blob/master/json.md#version).

leggomuhgreggo:
Thanks!

leggomuhgreggo:
Merged PR. Closing

Issue #2246, opened 2016-04-13
Urgent: current version out of sync with Bower, update git tags


implico:
The git version tag 1.5.9 leads to a wrong point, take look at the [line 54 in the repo](https://github.com/kenwheeler/slick/blob/7a79c43f2864ba4087e2961849c2da7bdcfe2de8/slick/slick.js#L54), and then install with Bower. You will get the old way (returns string vs. jQuery object):
```
customPaging: function(slider, i) {
  return '<button type="button" data-role="none" role="button" aria-required="false" tabindex="0">' + (i + 1) + '</button>';
},
```

**Solution**
Please run:
```
git tag 1.5.9 -f
git push origin 1.5.9 -f
```

**Edit:** additionally, the current minified version contains the old way, [ctrl+f "customPaging"](https://github.com/kenwheeler/slick/blob/7a79c43f2864ba4087e2961849c2da7bdcfe2de8/slick/slick.min.js), needs rebuilding

simeydotme:
@implico what are you talking about?
`customPaging` has been a string for a _long_ time (maybe always?) ... just taking a random tag like `1.3.9` you can see that it's been a string returned for ages: https://github.com/kenwheeler/slick/blob/9aaaf13cb6d7bb6162063fcfcd091bec6688fc0b/slick/slick.js#L58

I think what you're referring to is the `master` branch, which is the `HEAD` and hasn't been tagged for release yet. You're welcome to download the source from `master` until @kenwheeler has pushed a release

Please, also, don't spread bad practices by asking people to force a tag update... that could have all kinds of negative implications.

implico:
@simeydotme: yes, I was referring to the `HEAD`, so it would have been better to point that this should not be changed in the `1.*` version to remain backward compatibility (or handle if the returned value is a string/object).

I wasn't asking people to update the tag, just @kenwheeler :)

Anyway, nevermind.

PR #2557, opened 2016-10-10
Remove deprecated version field from bower.json file

wolfika:
refs #2555

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -6,7 +6,6 @@
         "slick/slick.less",
         "slick/slick.scss"
     ],
-    "version": "1.6.0",
     "homepage": "https://github.com/kenwheeler/slick",
     "authors": [
         "Ken Wheeler <ken_wheeler@me.com>"


leggomuhgreggo:
Hey @wolfika never having used bower, I wanted to get a better idea of the landscape in the hopes of heading off any subsequent complications merging this PR might invite. I remembered this discussion #2246 from a while back--doesn't look like it actually bears on this issue, but thought I'd reference it anyhow.

I guess my only remaining concern is how many people might be using on older versions of bower that rely on the version field declaration. Let me know if you have any thoughts on the subject. Definitely leaning towards merging this in.

Thanks

wolfika:
I would find it rational if Bower used git tags a default fallback for parsing a valid semver version if that could not be done from the `bower.json`. This is just a speculation, I don't know older Bower releases to that extent, but I would find that logical, some further research required.

PR #2576, opened 2016-10-18
Fix #2575

### LINKED ISSUE #2575: Prevent change slide to propagate to parent element
> When the carousel is inside a tag with a click event, clicking on an arrow or dot to change the slide will propagate to the parent element event.

###### `[  http://jsfiddle.net/natete/2L3m5pzg/2/  ]`  

1. ...  Click on an arrow or a dot

...  Just a slide change.

...  The parent click event is executed.

#### More Details

- Which browsers/versions does it happen on? All.
- Which jQuery/Slick version are you using? 1.6.0
- Did this work before? I don't know.

natete:
Add an event.stopPropagation to changeSlide handler to prevent click events to propagate to parent container.


leggomuhgreggo:
Excellent fix. Looks like some changes were made to .gitignore and ISSUE_TEMPLATE.md

I think the change to .gitignore makes sense, but should be in a separate commit, and I think the ISSUE_TEMPLATE.md changes might be by mistake?

Once that get's cleaned up I think we can get this merged.

Thanks @natete

julestruong:
Perfect, why isnt this already merged ?

webzlodimir:
Guys, please merge this, it is very useful fix

natete:
I removed the files I committed by mistake, I think you can proceed now with the merge.

PR #2647, opened 2016-12-05
Check hasOwnProperty before accessing field

hkrutzer:
I've been getting `Cannot read property 'breakpoint' of undefined` [here](https://github.com/kenwheeler/slick/blob/95e1be8/slick/slick.js#L1776) from our Javascript error tracking system. I have not been able to reproduce this outside of a testcase, and I am unsure what exactly is causing this error (perhaps some third party abusing the Array prototype), but this should solve it. 

Testcase [here](http://jsfiddle.net/ey086m7u/).

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1773,9 +1773,9 @@
             for ( breakpoint in responsiveSettings ) {
 
                 l = _.breakpoints.length-1;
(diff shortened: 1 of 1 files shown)


alex-ketch:
I was encountering the same issue and this fixes it.

PR #2720, opened 2017-01-23
making possible wrapping of dots with arrows

someson:
in case of cusomizing of controlling elements:
appendArrows: '.slider ~ .slider-controls',
appendDots: '.slider ~ .slider-controls',
it would create the elements in form of:
← o o o →
width-ajustable and without absolute positioning.


robertpro:
I would like this feature!

PR #2750, opened 2017-02-06
Fix for jQuery 3

eirslett:
https://api.jquery.com/ready/

> jQuery offers several ways to attach a function that will run when the DOM is ready. All of the following syntaxes are equivalent:
>
> $( handler )
> $( document ).ready( handler )
> $( "document" ).ready( handler )
> $( "img" ).ready( handler )
> $().ready( handler )
> As of jQuery 3.0, only the first syntax is recommended; the other syntaxes still work but are deprecated.


leggomuhgreggo:
Cool, thanks!

Issue #2821, opened 2017-03-29
Events not fired if # of slides less than or equal to 'slidesToShow'


flaviocysne:
When the number of elements/slides are less than or equal the value configured in slidesToShow settings's property, no arrows or dots are rendered and events are not fired.

Sample: 

====================================================================

###### `[  http://plnkr.co/edit/QqcVX7?p=preview  ]`  

use this Plunker to reproduce your bug: 
http://plnkr.co/edit/QqcVX7?p=preview

This sample is using an angular port of this component, but serves only as a proxy to this component's methods and attributes.

====================================================================

#### Steps to reproduce the problem

1. ...  any slick-carrousel sample set up that uses an slidesToShow's value greater than or equal to slide count.

====================================================================

#### What is the expected behaviour?

...  The expected behavior is that one bullet is shown and the click event is fired.

====================================================================

#### What is observed behaviour?

...  Nor the bullet is shown neither the click event is fired.

====================================================================

#### More Details

- Which browsers/versions does it happen on?
Chrome 42+
Firefox 3.4+

- Which jQuery/Slick version are you using?
//cdnjs.cloudflare.com/ajax/libs/jquery/1.12.4/jquery.min.js
//cdnjs.cloudflare.com/ajax/libs/slick-carousel/1.6.0/slick.min.js

- Did this work before?
No.

flaviocysne:
[PR accepted](https://github.com/kenwheeler/slick/commit/92c543900a9867551dff2c10e33296a977d6332c). Get the latest version of the [javascript component](https://github.com/kenwheeler/slick).

PR #2822, opened 2017-03-29
Fix bullet and click event problem when slideCount <= slidesToShow

flaviocysne:
Fix for the [Issue #2821](https://github.com/kenwheeler/slick/issues/2821)
This PR enable click events if number of slides is less than slidesToShow.
One bullet is shown when slideCount is lesser or equal than slidesToShow, also enabling currentSlide to be reset after changing its value.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -481,7 +481,7 @@
         var _ = this,
             i, dot;
 
-        if (_.options.dots === true && _.slideCount > _.options.slidesToShow) {
+        if (_.options.dots === true) {
 
             _.$slider.addClass('slick-dotted');
 
@@ -1335,7 +1335,7 @@
 
         var _ = this;
 
-        if (_.options.dots === true && _.slideCount > _.options.slidesToShow) {
+        if (_.options.dots === true) {
             $('li', _.$dots).on('click.slick', {
                 message: 'index'
             }, _.changeSlide);
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1039,10 +1039,14 @@
         var pagerQty = 0;
 
         if (_.options.infinite === true) {
-            while (breakPoint < _.slideCount) {
-                ++pagerQty;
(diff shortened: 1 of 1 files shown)


flaviocysne:
@kenwheeler any suggestion or comment? I really need this fix. At the moment I'm using my fork, but I still hope using yours.

pnewman8:
I also need this fix. In the short term I plan to use this fork. Thanks to flaviocysne.

My problem was that a call to slickGoTo was ignored (in slideHandler), because slideCount <= slidesToShow.

pnewman8:
On my call to slickGoTo, I had to pass the "dontAnimate" argument as true. Without that, only the selected slide was visible, at the left (it's actually the last of 3, with slidesToShow = 3). Haven't figured out why yet.

leggomuhgreggo:
Thanks for the PR. I definitely'd like to review this and get it merged. Would speed things along if someone could get the ball rolling on some testing. [Here's a codepen](https://codepen.io/leggomuhgreggo/pen/ONBMZV) with a bunch of different cases to make sure aren't being broken by the change -- just replace the slick.js with your fork ala rawgit or something.

nsarradin:
Thanks for the PR ! I have tested it on my project and it solves the slickGoTo method bug. Just a suggestion of improvement if I may, the infinite option doesn't seem to work when the number of slides <= slideToShow. It could be great to correct this :)

flaviocysne:
@pnewman8 could you publish a Code Pen sample or a Flickr sample that I can use to test against?

@leggomuhgreggo forked your pen and changed slideToShow values to the same number of slides per example. Let me know of any unexpected behavior, please. https://codepen.io/flaviocysne/pen/BRWmBK

@nsarradin what is the expected behavior when infinte is true and numer of slides <= slidesToShow? I didn't change it because I thought that arrows are used to change carrousel pages, not to change slides.

flaviocysne:
@kenwheeler is there anything still needed to accept this PR?

Issue #1937, opened 2015-12-01
Remove aria-live="polite" when slider is not focused


cshold:
`.slick-list` has `aria-live="polite"` on it, which tells screen readers to speak the content inside whenever it changes. This is great for the most part, but not ideal when you're not interacting with the slider. 

For example, imagine a screen reader scrolling down the page, past the slider, and still having the slide content read to them because `aria-live` exists on the element. It would be confusing to say the least.

I propose removing the aria attribute unless an element inside the slider is focused.

laneystroup:
I second the need for this; it is very annoying to have your actions constantly interrupted by the carousel if you are a screen reader user.

devben:
+1 
agreed this is very annoying for anyone using a screenreader

laneystroup:
FYI, I have arrived at a workaround that I am pretty happy with. If you disable autoPlay and then set initialSlide to a random number, then you avoid the problem of the carousel interrupting the screen readers, you still some the variety for revisits, and the users can cycle through the other pictures if they like. The tradeoff of course, is that if you really want the users to see the other slides then they may not see those. However, it is probably better to place that information elsewhere on the page anyway rather than hiding it inside a carousel. For my purposes, it works perfectly, as the slides are more of a background feature (just rotating pictures).

To implement this, just set initialSlide to "Math.floor(Math.random() * (n + 1))", where n is the number of slides.

Hope that helps!

depiction:
I agree that aria-live should be removed.

I tried the following to remove the attribute, then re-add it when the list has focus. Once focused on the slick list I waited until all slides had been rotated through. The attribute was added and removed appropriately, but the screen reader I use (VoiceOver) didn't read each slide as it changed. Since the live region isn't registered when added via JS the live-region can't be re-enabled on focus.

`			$slickList.removeAttr('aria-live');
			
			$slickList.on('focus', function() {
				$slickList.attr('aria-live', 'polite');
			});
			$slickList.on('blur').removeAttr('aria-live');`

dotherightthing:
+1 My client has four (!) Slick sliders on the page, so things are very noisy for us here. ;-)

edisch:
+1 On Chrome with VO for Mac, as a user tabs through each slide they receive audio feedback telling them the alt attribute of the slide image. However, on Firefox VO is not reading out the slides as they come through.

JordanDDisch:
PR to remove aria live #2843

classaxe:
[...]

PR #2843, opened 2017-04-14
1937: Removed aria live attribute

JordanDDisch:
As mentioned in this issue #1937 we do not need aria-live


leggomuhgreggo:
Thanks!

PR #2848, opened 2017-04-18
jQuery's 'ready' event deprecated

okonomiyaki3000:
The only acceptable way to handle the ready event in jQuery is now: 
`jQuery(function($) { /** whatever **/ });`


leggomuhgreggo:
Oh nice, a rare v2 PR. Honestly, I think we're trying to drop jquery altogether on v2, but just the same I'll merge this sucker in. Thanks @okonomiyaki3000

Issue #2847, opened 2017-04-18
Module loader doesn't check for "module"


matthew-dean:
Rollup (https://github.com/rollup/rollup) is a system for bundling ES6 modules. In the Slick code it has this:

```js
(function(factory) {
    'use strict';
    if (typeof define === 'function' && define.amd) {
        define(['jquery'], factory);
    } else if (typeof exports !== 'undefined') {
        module.exports = factory(require('jquery'));
    } else {
        factory(jQuery);
    }
```

The problem is that Rollup defines `exports` but does NOT define `module`. Therefore the code above fails. Is it possible to add a check for module before assigning an exports property?

leggomuhgreggo:
Man, module standards are one of the most stressful aspects of the JS community.

My guess is you could write and intermediary module that adapts the import standards. I'm not very familiar with Rollups orchestration, but I searched their issues and found [https://github.com/rollup/rollup/issues/578](this). Donno if you saw it or if it's actually relevant, but hopefully that helps.

Slick v2 will use ES6 modules, so there's a light at the end of the tunnel.

PR #2851, opened 2017-04-19
Check for "module" object (use single quotes)

matthew-dean:
Fixes https://github.com/kenwheeler/slick/issues/2847


leggomuhgreggo:
What's your level of confidence that this won't break other build systems? Do you have any documentation on the subject?

PR #2872, opened 2017-05-02
Center children if they fit in one slide (very rough)

dorik2k:
If there are no slides (children are all shown), set width relative to the number in percent so they all fill the entire track. Then the elements inside can be centred,

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1982,7 +1982,17 @@
 
         if (_.options.vertical === false && _.options.variableWidth === false) {
             _.slideWidth = Math.ceil(_.listWidth / _.options.slidesToShow);
-            _.$slideTrack.width(Math.ceil((_.slideWidth * _.$slideTrack.children('.slick-slide').length)));
+             
+            if (_.slideCount < _.options.slidesToShow) {
+                _.$slideTrack.width("100%");
+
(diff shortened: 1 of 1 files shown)


leggomuhgreggo:
Can you clarify the purpose of this? Thanks!

dorik2k:
First of all let me admit that the code is not very polished and does not take into consideration the configuration parameters.

What it does is spread the "slick-slide" div to fill the entire width of "slick-track" depending on slide number. For example, if there are 4 slides, and the "slidesToShow" is 4, all slides get a width of 25%. But, if there are 3 slides in the carousel, the slides get a width of 33%. As it is, the code sets the width only based on the "slidesToShow" variable and if there are fewer slides than the carousel ca fit, they are stacked in the middle. 

This can be done in the current version of the code by declaring the exact number of slides in the init code, but we had a situation where the number of carousels in the page, and also the number of slides in each carousel, is dynamic. It is complicated to declare each instance code dynamically for each slide. We initiated each carousel by looping each element and using the same parameters. If the carousel had more slides than, or equal to "slidesToShow", they would fit in the carousel as usual. But, if the carousel had fewer slides, they would spread and not get stacked in the centre of the container.

In the example included in the code, you have to set ".slick-slide { margin: 0; }" or the slides will wrap.

PR #2932, opened 2017-06-14
2639 corrected aria attributes on dots

### LINKED ISSUE #2639: Dots: aria-hidden true/false attr need to be inversed
In 1.6 the active-dot has aria-hidden="false" and the others aria-hidden="true". For screenreaders this means that they can not navigate by the dots.

Just inverse it:
- active dot: aria-hidden="true" + disabled on the button maybe good
- inactive dot: aria-hidden="false"

Thanks,
K

keeganstreet:
Fixes #2639


leggomuhgreggo:
Looks good. Thanks!

PR #2946, opened 2017-06-19
#1207 White spaces when using variableWidth and infinite to true

nicolaspernot:
Solve the #1207 issues about white spaces when using variableWidth and infinite to true, solution gave by xXanth0s (https://github.com/kenwheeler/slick/issues/1207#issuecomment-263908723)

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2341,7 +2341,7 @@
                         .attr('data-slick-index', slideIndex - _.slideCount)
                         .prependTo(_.$slideTrack).addClass('slick-cloned');
                 }
-                for (i = 0; i < infiniteCount; i += 1) {
+                for (i = 0; i < infiniteCount  + _.slideCount; i += 1) {
                     slideIndex = i;
                     $(_.$slides[slideIndex]).clone(true).attr('id', '')
                         .attr('data-slick-index', slideIndex + _.slideCount)


leggomuhgreggo:
Hey can you demo your changes in a pen, and then maybe test it against some of the existing functionality in [this test template](https://codepen.io/pen?template=ONBMZV)

Thanks!

nicolaspernot:
Here is it : https://codepen.io/nicolaspernot/pen/WOZwzQ

I added an example of a slider with variableWidth: true and  infinite: true.
Solution is clearly better but seems not perfect when slides width are not exactly the same as the sliders width. But better. I'll try to fix this last bug in a near future.

For the other sliders demos, I noticed no problem, except maybe some changing in green and pinky colors. But I think it's nothing else than CSS improvement.

leggomuhgreggo:
@nicolaspernot cool flashing is from the clone swap, and is known issue.

Did you want to go another round on this or you feelin a "perfect is the enemy of good" merge?

nicolaspernot:
@leggomuhgreggo I know I will correct the last issue because I work for someone who want it solved. He wants it solved, but it is not his priority. A day I will, for sure :)
So, I think it's for the better to merge now and let's create an issue for the remaining issue.

By the way, yes, i'm a perfect-is-the-enemy-of-good guy :)

leggomuhgreggo:
Thanks! Merged

PR #2952, opened 2017-06-23
Update dots alignment

doppiam:
I've noticed that every time i use slick (awesome plugin) the dots bar is not correctly aligned with the slider; figured out that a simple "left: 0;" will solve this problem!

Thanks for this awesome plugin!

### DIFF
--- a/slick/slick-theme.less
+++ b/slick/slick-theme.less
@@ -111,6 +111,7 @@
 .slick-dots {
     position: absolute;
     bottom: -25px;
+    left: 0;
     list-style: none;
     display: block;
     text-align: center;


ahmadalfy:
What do you mean by the dots aren't aligned? Can you create a fiddle demonstrating the issue you are trying to resolve?

doppiam:
I'm working on the fiddle to demostrate the issue! I will link it here as soon as i can;
Thanks for your reply.

doppiam:
Hi! I've done some fiddles, but the issue seems to not appear, here you can see what I'm talking about (I've encountered this problem in many project before, maybe I'm doing something wrong)
Thanks for the support!

schermata 2017-06-28 alle 11 46 55

PR #2958, opened 2017-06-28
Pull request commit for fixedSlideWidth option

stuartmcfarlane:
Hi.

I have a use case where the centerMode slide has a fixed width and the padding adjusts to maintain the fixed width.

Here is the jsfiddle test.

Best regards, Stuart


leggomuhgreggo:
Interesting. I'd like to avoid adding new options if possible and I think this fits decently within the scope of center padding.

What do you think about refactoring to support calc eg `calc((100% - $fixed-width) / 2)`?

PR #1901, opened 2015-11-18
Correctly center slider

niaccurshi:
Ensure the slides are centered if slideCount is less than slidesToShow.

(This is only the unminified version, haven't the time to do anything other than fix this issue!)

PR #2966, opened 2017-06-30
Aligns slides in the center

corvus-007:
Aligns slides in the center if the number of slides is less than slidesToShow.
https://codepen.io/corvus-007/pen/LLdPqx?editors=1100

### DIFF
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -55,6 +55,8 @@
     left: 0;
 
     display: block;
+    margin-left: auto;
+    margin-right: auto;
 }
 .slick-track:before,
 .slick-track:after
--- a/slick/slick.less
+++ b/slick/slick.less
@@ -44,6 +44,8 @@
     left: 0;
     top: 0;
     display: block;
+    margin-left: auto;
+    margin-right: auto;
 
     &:before,
     &:after {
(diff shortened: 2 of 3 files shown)


tkuther:
Looks related to #1901

leggomuhgreggo:
Couldn't think of any instance where this'd mess things up. Looked good in my testing. My only hesitancy is adding styles that might (in some unknown case) conflict, entropy yada yada, but it's the theme file which is optional, so I'm going to green light it. 

Thanks for the PR

mfeltscher:
This breaks existing setups with `centerMode: false` and a small number of slides which should be left aligned.

DanielRuf:
> This breaks existing setups with centerMode: false and a small number of slides which should be left aligned.

Exactly. Also by default this was not centered before so this is a breaking change.

warudin:
Since 1.7 sliders with a small number of slides are lined out in the center, but in various cases this isn't the desired effect. What setting do I use to prevent this?

DanielRuf:
> Since 1.7 sliders with a small number of slides are lined out in the center, but in various cases this isn't the desired effect. What setting do I use to prevent this?

You have to reset the margins.

PR #2969, opened 2017-07-03
conditional @font-face hack for slick-theme.less

vanilla-thunder:
font files should only be loaded if they are needed, like sass file already does.


leggomuhgreggo:
Good call. Merging.

PR #2981, opened 2017-07-11
Removed Aria-pressed label for html validation reasons

tbirdsall:
In order to pass HTML5 validation (https://validator.w3.org/nu) and be fully WCAG AA compliant, the Aria-pressed label was removed, as it is not allowed on an li element at that point. 

Codepen for test cases https://codepen.io/tbirdsall/pen/PjXwGW

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2941,20 +2941,12 @@
             _.$dots
                 .find('li')
                     .removeClass('slick-active')
-                    .end()
-                .find('button')
-                    .attr({
-                        'aria-pressed': false
-                    });
+                    .end();
 
             _.$dots
                 .find('li')
                 .eq(Math.floor(_.currentSlide / _.options.slidesToScroll))
-                .addClass('slick-active')
-                .find('button')
-                .attr({
-                    'aria-pressed': true
-                });
+                .addClass('slick-active');
 
         }
 


leggomuhgreggo:
Hey thanks for the PR and especially for including a demo. Can you link to an article discussing this? My recollection is that the w3 validator isn't always 100%, and I wonder if, since it applies to `button` instead of `li`, it might just be an issue with the validator rather than actual compliance.

Thanks

cappadona:
I believe the issue is that the `aria-pressed` state is [valid only for the `button` role](https://www.w3.org/TR/wai-aria/states_and_properties#aria-pressed), and Slick's current use of the `tab` role leads to a failure of [WCAG2.0 4.1.2](https://www.w3.org/WAI/WCAG20/quickref/#qr-ensure-compat-rsv)

leggomuhgreggo:
Cool thanks for the clarification and the links @cappadona.

I think this adds up to me. @cielt @jasonday @scottaohara this jive with y'all?

jasonday:
@leggomuhgreggo, I think @cappadona is correct - `aria-pressed` does not apply/ behave as intended due to other roles in play.

leggomuhgreggo:
Cool thanks for weighing in @jasonday.

I'm going to merge this sucker. Thanks for making the PR @tbirdsall

cielt:
Nice edit @tbirdsall 👍  Sorry, just catching up with this, but to add to the rationale for removing `aria-pressed` here: it is an attribute that mainly applies to buttons serving as a toggle (with 2 states) whereas, in Slick, the buttons are serving as part of a UI component that doesn't really leverage them as such: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role#Toggle_buttons

PR #2983, opened 2017-07-11
Added slide parent property to support using an existing slide track.

### LINKED ISSUE #212: list based  slider
From the time i've spent so far with the plugin I think its great and very flexible. I have however run into an issue which, whether this is or not.

It doesn't seem that one can create a slider based upon an unordered list:

```
<ul>
    <li>item one</li>
    <li>item one</li>
    <li>item one</li>
</ul>
```

Is there a reason for this? Would be nice if markup didn't matter.

johnrom:
closes #212 

Added code to keep track of whether the `_.$slideTrack` was originally the parent of the slides (existed previously) or was created during the slick buildout process. Allows one to use `ul` as a `.slick-track` and `li` as a `.slick-slide`

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -81,6 +81,7 @@
                 swipeToSlide: false,
                 touchMove: true,
                 touchThreshold: 5,
+                track: false,
                 useCSS: true,
                 useTransform: true,
                 variableWidth: false,
@@ -107,7 +108,10 @@
                 scrolling: false,
                 slideCount: null,
                 slideWidth: null,
+                $slideParent: null,
                 $slideTrack: null,
+                slideTrackIsParent: false,
+                $slideTrackOriginalParent: null,
                 $slides: null,
                 sliding: false,
                 slideOffset: 0,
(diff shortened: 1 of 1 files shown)


leggomuhgreggo:
Hey I think the feature is legit. Anyway to reduce the number of new options to one? Thanks

johnrom:
There is only one new option, but there are three new things that need to be kept track of internally. I can reduce it to two by checking if `_.$slideParent === _.$slideTrack` each time it is referenced, but I chose to keep an extra boolean `slideTrackIsParent` as it more accurately captures the original state of the slider in the event that _.$slideTrack changes somehow.

leggomuhgreggo:
It's a minor thing, but reducing the number of keys relating to a single feature would help readability, so it'd be cool if that were possible. Thanks

johnrom:
It's been removed

leggomuhgreggo:
Hmm, I'm still seeing three additional options. Apologies for the difficulty, but do you think you could factor to just use one? 

Thanks

johnrom:
There is now one additional option and one internal property. The internal property tracks the distinction between $slider, $slideTrack and which one is really the parent of the slides before and after slide creation, so that elements can be properly returned to their initial position after using `unslick`. I don't have the time to attempt any sort of refactor that would remove this property because it is fundamental to the entire change, and I do not plan to start over from scratch.

leggomuhgreggo:
Cool, I'll try to test this later, and then hopefully merge it in. Thanks.

johnrom:
FYI this is not pending author response anymore, and I've updated it with `master`

Lukas238:
Any progress on this feature?
Best regards!

tomkyser:
any updates regarding this?

Shioku1337:
I would also be interested in this feature being added.

PR #2992, opened 2017-07-18
Replacing constant window/document wrapping with single wrapped reference

nominalaeon:
Pedantic, asinine, triflin', "who the f*ck does this guy think he is." These are just a few of the slurs used to describe this pull request.

And yet, here we are.

PROMPT: I was trying to track down where `drag` gets its threshold from and noticed `window` and `document` were getting fresh JQuery wrappers with every use. It made me curious more than anything so I began digging in to see what, if any, performance gains there were to referencing a single, global instance of these instead of rewrapping them every time.

CHANGES: I set `const` declarations at the top and used them to replace all instances of `$(window)` and `$(document)`.

PAIN-POINT: JQuery does a great job of detecting if `window` is the object getting wrapped and, as far as I can determine, not rewrapping it. I can't tell if it's doing this for `document` or not. Either way, it has to do logic to make that determination. Logic that we can just do from our end to help it along.

Additionally, [the performance gains that I could find](https://jsperf.com/caching-jquery-selectors) seem more relevant to looped instances than seldom used one-offs, so really this comes down to a matter of style/opinion. 

My enthusiasm isn't nearly as high as my curiosity was, so I'm not pushing this issue too hard. If I hadn't noticed that work was being done on a completely fresh version, I probably wouldn't be PR'ing at all, but this seems like a good time to make this update if it needs to get made at all.

I made this test with version 1.6.0: https://codepen.io/nominalaeon/project/editor/Dnnxzj/#0

I've also implemented the change in a microsite that is doing some heavy lifting slider stuff using angular-slick and nested sliders. Here are some screenshots:

slider-1
slider-2

In summation, gentlemen, I propose you accept this change and embrace it in your hearts. As true believers of JQuery it's our job, nay our Bob-given duty, to use it responsibly and set good examples for the children.

Or not, whatever.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -3,6 +3,9 @@
 
 const $ = window.$ || window.jQuery;
 
+const $window = $(window);
+const $document = $(document);
+
 export default {
   cleanUpEvents() {
     if (this.options.dots && this.$dots !== null) {
@@ -27,7 +30,7 @@ export default {
 
     this.$list.off("click.slick", this.clickHandler);
 
-    $(document).off(this.visibilityChange, this.visibility);
+    $document.off(this.visibilityChange, this.visibility);
 
     this.$list.off("mouseenter.slick", $.proxy(this.setPaused, this, true));
     this.$list.off("mouseleave.slick", $.proxy(this.setPaused, this, false));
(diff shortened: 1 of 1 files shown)


nominalaeon:
Sidenote, if anyone can ELI5 drag's threshold, I would be forever indebted.

leggomuhgreggo:
Hey thanks for putting in the effort on this. Not opposed to putting window/document in variables, but I think we're still trying to support es5, so gonna have to swap out the const for var.

I think touch threshold refers to the number of pixels required before a slide change occurs. See [here](https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L2722-L2728)

Thanks

nominalaeon:
Thank you for the direction on touch threshold, that is exactly what I was hunting for.

I used `const` to play along with the `const $ = window.$ || window.jQuery;` declaration above it. I haven't had a chance to poke around too much, but the Webpack config looks like it's using Babel, so the es6 stuff should get distilled into compatible biz in the end. Happy to change it if that's not the case.

Thanks for looking this over, legg!

leggomuhgreggo:
Ohhhh, this is the v2 branch, cool in that case you're fine with const.

nominalaeon:
Nice, thanks! Updating version 1 with this seemed more risk than reward

PR #2996, opened 2017-07-20
Fix Variable Width + Infinite (#1207)

MattStopa:
This fixes Variable Width + Infinite (#1207)

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2560,6 +2560,26 @@
             return;
         }
 
+        var slickWidth = _.$slideTrack.parent().width()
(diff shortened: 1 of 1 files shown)


leggomuhgreggo:
Cool, only thing this needs is some test demos. Here's this [codepen template](https://codepen.io/pen?template=ONBMZV) you can use, just replace with your branch

daverdalas:
@MattStopa Your solution works incorrectly when slides use margins. Maybe you should you use .outerWidth() instead of .width()?

pSchaub:
@MattStopa Can you fix the question from @daverdalas and request a merge?

Issue #1871, opened 2015-11-06
Editing Row amounts for responsive


Fionah:
I have been using slick slider a lot and I love it! I am having an issue with the rows for my responsive settings. I want two rows for desktop which then goes to one row for a smaller screen, the responsive settings recognizes the changes to slide amounts but not rows. The code I have used is below, can you tell me where I am going wrong:

                $('.ptype-slick').slick({
                    infinite: true,
                    arrows: false,
                    rows: 1,
                    slidesToShow: 1,
                    mobileFirst: true,
                    responsive: [
                        {
                            breakpoint: 360,
                            settings: {
                                slidesToShow: 2,
                            }
                        },
                        {
                            breakpoint: 450,
                            settings: {
                                slidesToShow: 3,
                            }
                        },
                        {
                            breakpoint: 600,
                            settings: {
                                slidesToShow: 4,
                            }
                        },
                        {
                            breakpoint: 700,
                            settings: {
                                 rows: 2,
                                slidesPerRow: 5,
                }
            }
                    ]
                });

Many thanks in advcance!
Fiona

lukasz-virtualo:
Same thing with going with the responsive things in the reverse (large desktop as the default and then smaller screens later on):
```js
{
    autoplay: true,
    autoplaySpeed: 3000,
    pauseOnHover: false,
    arrows: false,
    dots: false,
    infinite: true,
    speed: 300,
    cssEase: 'linear',
    rows: 2,
    slidesPerRow: 6,
    responsive: [
        { breakpoint: 1024, settings: {
            rows: 1,
            slidesToShow: 5,
            slidesToScroll: 5
        }},
        { breakpoint: 640, settings: {
            rows: 1,
            slidesToShow: 4,
            slidesToScroll: 4
        }}
    ]
}
```
Upon resize to the smaller screen I get a very small grid of 6 on 2 elements occupying perhaps a 20% of the available area in the container.
Using a single row for all resolutions works OK.

Lysande:
I'm currently having the exact same problem, both with and without mobileFirst enabled. 

The row count specified in the main settings are the ones that stick, any rows specified on a breakpoint are ignored. It appears that slidesPerRow are also ignored.

AlexeyKarash:
+1 have same problem now.

hakastein:
+1 same bug

kenwheeler:
Can someone please jsfiddle this

hakastein:
[...]

PR #3001, opened 2017-07-25
Fixes bug in responsive rows

amapparat:
Open since 2015: https://github.com/kenwheeler/slick/issues/1871

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -558,7 +558,7 @@
         newSlides = document.createDocumentFragment();
         originalSlides = _.$slider.children();
 
-        if(_.options.rows > 1) {
+        if(_.options.rows > 0) {
 
             slidesPerSection = _.options.slidesPerRow * _.options.rows;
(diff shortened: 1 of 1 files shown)


leggomuhgreggo:
Cool, only thing this needs is some test demos. Here's this [codepen template](https://codepen.io/pen?template=ONBMZV) you can use, just replace with your branch

danjaywing:
I'm concerned as myself and two others raised the above 'fix' as breaking for other sliders on the same page. I assume this has been tested in that use case?

nyanginator:
@danjaywing I concur. The fix doesn't really make sense to me. Correct me if I'm wrong, but I believe _.options.rows isn't ever supposed to be <= 0. How else would it distinguish between single-row and multi-row? Should those code blocks really be getting executed in single-row mode?

Liroo:
This PR made slick unusable, going totally in favor of [swiper](https://github.com/nolimits4web/swiper/)...

PR #3030, opened 2017-08-03
Add a margin:auto on the img of the slick slide of  Slick.scss

TheodoreLaurent:
That margin: auto would have saved me at leat two hours ;-) sure I'm not the only one who struggled!
All in all, great work on that Slick thing, think it's very good.

### DIFF
--- a/slick/slick.scss
+++ b/slick/slick.scss
@@ -70,6 +70,7 @@
     }
     img {
         display: block;
+        margin: auto;
     }
     &.slick-loading img {
         display: none;


leggomuhgreggo:
Hey @TheodoreLaurent thanks for the PR. Definitely see the logic of this, but the slick.css/scss/less is intended to be super unopinionated. I could see this being good in slick-theme.css/scss/less, but even then, I worry about people fighting slick styles in cases where not every image is intended to be centered. If I see a lot of clamor for this I'd definitely be open to merge, but I'm going to table for the time being. Thanks!

TheodoreLaurent:
Hey, thanks for your reply,
totally understand although in my opinion most people want their images centered rather than not ;-)

Peace man

--
Theodore Laurent
+44 7495 668462

On 3 Aug 2017, at 19:06, Greg Westneat <notifications@github.com<mailto:notifications@github.com>> wrote:


Hey @TheodoreLaurent<https://github.com/theodorelaurent> thanks for the PR. Definitely see the logic of this, but the slick.css/scss/less is intended to be super unopinionated. I could see this being good in slick-theme.css/scss/less, but even then, I worry about people fighting slick styles in cases where not every image is intended to be centered. If I see a lot of clamor for this I'd definitely be open to merge, but I'm going to table for the time being. Thanks!

PR #3032, opened 2017-08-04
Added option to enable focus on slide change

### LINKED ISSUE #3020: "Accessibility: true" causes scroll jump to slick on click and viewport changes
Good morning and sorry for my english.

After the latest updates (now I have 1.7.1), clicking on slides or change the viewport using the "breakpoint" property, let the browser jump with a scroll at the slick position.
Is the same issue posted here some V. ago and marked resolved https://github.com/kenwheeler/slick/issues/1537.

The only fix  found is to enter the parameter "accessibility: false".

On I.E Edge, Chrome, Firefox.

Thanks!

leggomuhgreggo:
There've been issues about the jump resulting from focus being applied after slide change.

Cases that've been identified as problematic:
* Slick methods that change the slide
* Slides that are partially obscured by the top of the browser

I've (reluctantly) added a new option called `focusOnChange` which defaults to `false`, and updated the docs.

In order to achieve the a11y focus after change, this setting needs to be enabled.

Demos:
[Before](https://codepen.io/leggomuhgreggo/pen/ZJBBGO) & [After](https://codepen.io/leggomuhgreggo/pen/RZooNY)

Closes #3020

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -20,8 +20,8 @@ to serve the files as close, and fast as possible to your users:
 ##### Example using jsDelivr
 
 Just add a link to the css file in your `<head>`:
-```html
 
+```html
 <link rel="stylesheet" type="text/css" href="//cdn.jsdelivr.net/gh/kenwheeler/slick@1.7.1/slick/slick.css"/>
 <!-- Add the slick-theme.css if you want default styling -->
 <link rel="stylesheet" type="text/css" href="//cdn.jsdelivr.net/gh/kenwheeler/slick@1.7.1/slick/slick-theme.css"/>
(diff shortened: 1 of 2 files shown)


leggomuhgreggo:
@cielt Hey wanted to loop you in on this. I know it's not the most appetizing prospect from the perspective of an a11y stalwart, but it's become pretty clear that it's an issue. Let me know if you have any thoughts, and, as always, thanks!

cielt:
@leggomuhgreggo: solid fix from my perspective.  totally appreciate that setting `focus` is an opinionated and potentially disruptive default people might not want the visual effects of depending on use case.  thanks for resolving this and updating the docs! 👍

cielt:
(apologies -- should have included this in the prev note but was having connectivity issues)

@leggomuhgreggo: just a couple of small follow-up thoughts / questions:

1. any plans to do something with `focusOnSelect`?  i think with `focusOnChange` the a11y concern is covered; selecting (if this means clicking or otherwise interacting with an element that is able to receive `focus`) will by default transfer browser focus to that element (e.g., a slick dot or arrow button).  Implementing this change of `focus` programmatically is only necessary where focus is not transferred by the browser automatically.  My suggestion would be to remove unless you have other designs for this;
1. we can remove the comment, around line `1700`, `// for non-autoplay: once active slide (group) has updated, set focus on first newly showing slide`; 
1. agree with the `README` update for the `accessibility` option referencing `focusOnChange`; if it may help, i would suggest adding a note to the `focusOnChange` description, explaining that this is so that screen readers are prompted to call attention to the slide newly showing;
1. depending on what you decide in `1` above, description of `focusOnSelect` may not be needed.

Let me know if any questions.  And thanks!

leggomuhgreggo:
Cool, I'll look into deprecating focusOnSelect, and in the meantime remove it from the docs (it was absent before this PR, as a matter of fact). Also removed that comment. Thanks @cielt

PR #3075, opened 2017-08-24
Added shuffle slides feature (ability to exclude first)

midnight-studios:
Shuffles slick slides before slick gets initialized with the option to include or exclude the first slide.

[Demo](https://jsfiddle.net/midnightstudios/efjqnqgz/)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -68,6 +68,8 @@ Example:
 
 Option | Type | Default | Description
 ------ | ---- | ------- | -----------
+shuffle |  boolean | false | Shuffles slick slides before slick gets initialized.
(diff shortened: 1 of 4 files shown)


leggomuhgreggo:
Hey appreciate the PR, but I'm not sure the feature has enough demand/added-value to justify the entropy. Will definitely keep it in mind. Thanks!

midnight-studios:
Hey. Thank you for the feedback. Just for a use case example, when using the slider to show products or brands this feature helps to mix it up so that the same products are not loaded (in the same order) on each load/initialization. This helps to give slides 'fair' exposure.

Issue #3009, opened 2017-07-27
Reimplement Show dots only when needed


cbieser:
In 1.6.0 Dots where only inserted if needed

```
  Slick.prototype.buildDots = function() {
    var _ = this,
      i,
      dot;

    if (_.options.dots === true && _.slideCount > _.options.slidesToShow) {
      _.$slider.addClass('slick-dotted');
```

Please reimplemet this or make it configurable.

Version 1.7.0

leggomuhgreggo:
Cool, would you like to make a PR?

SRTM:
+1 Please re-implement this.

cbieser:
@see #3080

leggomuhgreggo:
Merged into master. Thanks!

yangkennyk:
dots seem to show when there is only 1 slide again in 1.8

juliussveikauskas:
I found there is that logic in slick.js. But if I use slick.min.js it doesn't work. So if you need this feature you can try to use slick.js instead of slick.min.js.

octipus:
> I found there is that logic in slick.js. But if I use slick.min.js it doesn't work. So if you need this feature you can try to use slick.js instead of slick.min.js.

Thank you! This worked for me. I had to manually minifiy the .js file but that's fine

PR #3080, opened 2017-08-28
Show dots only when needed

cbieser:
Issue #3009: Show dots only when needed


leggomuhgreggo:
Cool thanks @cbieser 

Related: #2822

PR #3093, opened 2017-09-02
Update jsDelivr CDN links

MartinKolarik:
First, big congrats - [slick receives more than 2,6 B requests per month at jsDelivr](https://www.jsdelivr.com/package/npm/slick-carousel), which makes it the most popular package we host!

While our new back-end supports serving files from both npm and GitHub, we recommend always using only one of these methods for every package, with npm being the preferred one (using only npm means you can get detailed package usage statistics and npm packages are also searchable on our website).

I updated the links to use our npm endpoint. I also noticed you don't provide minified CSS files in your package. jsDelivr can now do on-demand minification, which means you could link to the following files for minified versions:
 - https://cdn.jsdelivr.net/npm/slick-carousel@1.7.1/slick/slick.min.css
 - https://cdn.jsdelivr.net/npm/slick-carousel@1.7.1/slick/slick-theme.min.css

Let me know if I should update the links to use minified CSS files instead.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -22,15 +22,15 @@ to serve the files as close, and fast as possible to your users:
 Just add a link to the css file in your `<head>`:
 
 ```html
-<link rel="stylesheet" type="text/css" href="//cdn.jsdelivr.net/gh/kenwheeler/slick@1.7.1/slick/slick.css"/>
+<link rel="stylesheet" type="text/css" href="//cdn.jsdelivr.net/npm/slick-carousel@1.7.1/slick/slick.css"/>
 <!-- Add the slick-theme.css if you want default styling -->
-<link rel="stylesheet" type="text/css" href="//cdn.jsdelivr.net/gh/kenwheeler/slick@1.7.1/slick/slick-theme.css"/>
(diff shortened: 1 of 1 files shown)


leggomuhgreggo:
Hey @MartinKolarik thanks! When we switched to 1.7.1, I tried just updating the URL in the readme and the links seemed to break -- is there something I should do for the new release to make sure the new version is propagating to jsdelivr?

MartinKolarik:
@leggomuhgreggo 1.7.1 was the first version released after we switched to the new `/npm/` and `/gh/` endpoints - the old links (using the format `/project/version/file`) stopped updating at that time, which is why 1.7.1 didn't work.

The advantage of our new system is that it's fully automated and everything is available instantly after publishing, so there's no more work required from you.

leggomuhgreggo:
Ah so when the npm updates, the new links will work?

leggomuhgreggo:
Or oh -- the new link format in your PR will work when the GH updates too?

MartinKolarik:
The new links in this PR point to npm and have all npm versions (even the old ones) available immediately.

leggomuhgreggo:
Cool, I still need to work out the NPM publishing with Ken, but when that happens I'll update the links. Thanks!

PR #3097, opened 2017-09-05
Changed the logic for adding aria describedby attribute

tbirdsall:
There was a small bug where if the number of slides in a carousel is less than or equal to the slidesToShow variable, the aria-describedby attribute was still being added (with the next/prev button ID) even though the next/prev buttons are not in the DOM at that point.

This fix just checks to see if the corresponding button exists before adding the aria-describedby attribute.

Codepen for test cases https://codepen.io/tbirdsall/pen/PjXwGW

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1322,9 +1322,12 @@
                 });
 
                 if (slideControlIndex !== -1) {
-                    $(this).attr({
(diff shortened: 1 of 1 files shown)


leggomuhgreggo:
Excellent work

Issue #826, opened 2014-11-24
Selecting/Highlighting text


asnoba:
Hi,

I'm not sure whether this is a feature request or a bug, but as far as I can tell selecting/highlighting text in the active slide isn't possible – even if the `draggable` option is set to `false`. I think a fix for this would be nice for accessibility reasons. Any ideas?

I've included an example here: http://codepen.io/asnoba/full/yyNOQj/

BR,
Asger

jmcglone:
Agreed and +1 for setting `user-select: none` when `draggable` is `true`

I see this was added and then removed from milestone 1.4–any plans to add to future milestones?

safareli:
why was this removed?

reloaded:
@simeydotme  Any idea if this feature is going to be implemented? Why was it decided to not implement this (or not document it)?

nanastya:
Issue still exists. I'm sure a lot of developers would really appreciate if it was solved finally. Please, kenwheeler, help us out :)

lingtalfi:
+1

skyyip:
First Set the option `draggable: false` ,
Then override the css:
```
.slick-slider {
  -webkit-user-select: text;
  -moz-user-select: text;
  -ms-user-select: text;
  user-select: text;
}
```

pedroadame:
+1

PR #3112, opened 2017-09-15
User-select: none only when draggable: true

pedroadame:
PR for #826 to make text selectable unless `draggable: true` (default).

Demo here. All sliders have `draggable: false` and text is selectable without affecting behavior. https://codepen.io/pedroadame/pen/oGXoxR

### DIFF
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -6,10 +6,10 @@
     display: block;
     box-sizing: border-box;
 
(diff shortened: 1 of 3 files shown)


ricardovigatti:
Is this already planned to be merged?

pedroadame:
After more than one year, I still don't know. /shrug

PR #3118, opened 2017-09-17
Fix row unwrapping #3117

### LINKED ISSUE #3117: Changing number of rows breaks the layout
Changing number of rows will cause layout to break because slick fails to remove the wrappers added for rows when switching back to rows = 1.

In the example below the original default settings has `rows: 2`, but it will be changed to `rows: 1` by `breakpoint` rule.

https://jsfiddle.net/1aktqwy6/2/

1. set the default number of `rows` to something greater than 1
2. set a breakpoint where number of rows is 1

The slider should respect the settings coming from breakpoint.

The slider will display more rows instead of one because the wrapping used for rows is not removed when break-point is being applied

#### More Details

- Which browsers/versions does it happen on?
Every browser
- Which jQuery/Slick version are you using?
Latest, 1.8.0
- Did this work before?
Don't know.

Luxian:
This is probably not the best way to achieve this, but I'm open to feedback. I didn't know how to generate the minified version so it's untouched.

Fixes #3117

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -822,7 +822,7 @@
 
         var _ = this, originalSlides;
 
-        if(_.options.rows > 0) {
+        if(_.$slider.data('slick_rows_wrapped')) {
             originalSlides = _.$slides.children().children();
             originalSlides.removeAttr('style');
             _.$slider.empty().append(originalSlides);
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -582,6 +582,7 @@
             }
 
             _.$slider.empty().append(newSlides);
+            _.$slider.data('slick_rows_wrapped', true);
             _.$slider.children().children().children()
                 .css({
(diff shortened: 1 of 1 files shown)


slick/slick.js
```diff
@@ -822,7 +822,7 @@
 
         var _ = this, originalSlides;
 
-        if(_.options.rows > 0) {
+        if(_.$slider.data('slick_rows_wrapped')) {
```
Luxian: `_.options.rows` is already updated to new settings when it gets to this point, so there is no reliable way to check if the rows wrappers where added or not other than saving them somewhere else than `options`.

Luxian:
This apparently was fixed in the master branch by checking against `_.options.rows > 0` and my problem doesn't show up anymore. This patch is not needed if you use the latest version.

PR #3146, opened 2017-10-03
Make package name match repository name

bheatonf1:


kenwheeler:
I think it should be kept slick-carousel so that its consistent across bower and npm. slick is taken on npm.

Issue #3168, opened 2017-10-16
Outline: none makes slider inaccessible by keyboard


eladrin201:
> Outline: none is bad for accessibility and needs to be removed.

Open issue/duplicate issue: https://github.com/kenwheeler/slick/issues/3131
(I'm opening a new one to follow your guidelines)

====================================================================

###### `[  http://jsfiddle.net/eladrin201/1yb09ckc/1/  ]`

To view the issue, make sure you click into the html panel and hit tab. 

====================================================================

#### Steps to reproduce the problem

1. Click on the page or pane for slider. Hit tab.
2. Notice that the outlines are not visible. Users who rely on keyboard as a tool to navigate, can't keep track of where they are. 

====================================================================

#### What is the expected behaviour?

Maintaining outlines are important for users who use the keyboard to navigate to know where they are on the site. 

**Importance of outlines:**
* [a11yproject on outlines](http://a11yproject.com/posts/never-remove-css-outlines/)
* [WebAIM on outlines](https://webaim.org/blog/plague-of-outline-0/)

Second issue: Dots are not tabable due to tab index. 

====================================================================

#### What is observed behaviour?

Hitting tab, the user can actually get to buttons, slide content, etc, but they can easily get lost since the outline of focus is hidden or 0. Also you cannot tab to each dot rendering the inactive dots inaccessible by keyboard.

====================================================================

#### More Details

- Which browsers/versions does it happen on?
- Which jQuery/Slick version are you using?
- Did this work before?

This happens across all browsers (chrome, firefox, safari) and is primarily a style issue. 

I am using the most recent versions of jQuery and Slick.

eladrin201:
I made this fix, was pretty simple. Here is the PR: https://github.com/kenwheeler/slick/pull/3169

Jakobud:
This is an important PR. Is slick no longer updated?

eladrin201:
Good question. I would like to see this get merged in. We use Slick as an optional accessible slider, but I have to modify the style sheets for compliance. Is there any updates I need to make to my PR?

PR #3169, opened 2017-10-16
updating styling four outline and adjusting focus on the dots to corr…

eladrin201:
updating styling four outline and adjusting focus on the dots to correctly reflect the same one as the slide


eladrin201:
Fix for issue #3168

PR #3191, opened 2017-10-25
add unpkg.com cdn url

kukac7:


README.markdown
```diff
@@ -16,6 +16,7 @@ to serve the files as close, and fast as possible to your users:
 
 - https://cdnjs.com/libraries/slick-carousel
 - https://www.jsdelivr.com/projects/jquery.slick
+- https://unpkg.com/slick-carousel
```
dextel2: ..and minified?

dextel2: ok just saw after commenting that default link redirects to non-minified version

kukac7: https://unpkg.com/slick-carousel/slick/slick.min.js

Issue #3264, opened 2017-12-13
Slide width is 1 pixel too large on odd resolutions


hlohrenz:
I'm designing a carousel that displays two products and when you click the right arrow, it shows the next two products.

However, on odd resolution screens, Slick.js seems to round UP on it's width calculation per slide when it should round down in my opinion. The reason for this is because I have a border around my product images.

So when my container is 519px in width, it makes the slides 260px.. this will equal 520px in width making my last slide get 1px cut off.. You can see the images below:

![slickjs-1pxtoolg](https://user-images.githubusercontent.com/34523008/33962588-0d19c94e-e020-11e7-8b1f-b96ccce90ad2.PNG)

This is causing the border on the last product image to be cut off.. I looked through the code and found an issue in the setDimensions method.. I feel like it should use floor, not ceil:

`_.slideWidth = Math.ceil(_.listWidth / _.options.slidesToShow);`

Can anybody fill me in on this or how do I go about making changes?

Trystanr:
It's because your resolution has an odd number of pixels. You can't show half a pixel.

hlohrenz:
Uhhh... I get that. That's why I mentioned it in the issue? You can't always assume a user will be viewing your website on an even number browser width...
This is also why I suggested using floor instead of ceil.

PR #3269, opened 2017-12-18
Suggesting round down so the entire slide shows on odd resolution screens.

hlohrenz:
#3264 in the SetDimensions method, we should probably round down instead of up so 1 pixel isn't cut off on odd resolution screens.


Haydzyo:
I was having the same issue. I resolved it by changing all of the ceil() calls to floor() (In your PR you've only changed one, line 2052 was required as well :)).

PR #3368, opened 2018-03-11
Allow slidesToScroll to work with swipeToSlide

danielkorte:
I needed slidesToScroll to work with swipeToSlide. This fiddle shows how enabling swipeToSlide causes the carousel to always use 1 for slidesToScroll:
http://jsfiddle.net/rmd0bczg/1/

My updated version allows swipeToSlide to work along with slidesToScroll:
http://jsfiddle.net/mq7chax1/6/


terraelise:
Thank you!

PR #3397, opened 2018-04-04
PATCH: Allows adaptiveHeight option with multiple slides shown.

bakerej:
The current carousel implementation of adaptiveHeight only works for single-slide carousels. [[fiddle](http://jsfiddle.net/ebaker/zvxmd8Ld/3/)]

This patch allows for multiple slides, adapting to the height of the tallest item. [[fiddle](http://jsfiddle.net/ebaker/h5v0dppe/15/)]

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -246,8 +246,17 @@
 
     Slick.prototype.animateHeight = function() {
         var _ = this;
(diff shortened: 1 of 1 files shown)


apmeyer:
Thanks for this solution. Going with it for now. Would love to see this merged.

Issue #3268, opened 2017-12-18
Multiple accessibility issues and solutions to fix


goetsu:
> short description of the bug / issue, provide more detail below.
This is an issue about accessibility problems that are due to slick code structure and behaviors.
====================================================================

.slider : 
* remove role="toolbar" because it's not relevant here. Sliders are not toolbars see https://www.w3.org/TR/wai-aria-1.1/#toolbar 

.slick-prev, .slick-next : 
* remove role="button" as you already use button element, 
* when slider on infinite loop mode content inside prev button on first slide need to be "Go to last slide" otherwise user will not know it's looping
* when slider on infinite loop mode content inside next button on last slide need to be "Go to first slide" otherwise user will not know it's looping
* By default I recommend to use "Previous slide" and "Next slide" as it's more understandable that just "Previous" and "Next"
* By default I recommend to move focus on the new active slide (or first new one in case of multiple items) when user activate .slick-prev / .slick-next specially when there is focusable elements inside active slide

.slick-list : 
* remove aria-live at least when autoplay is one otherwise it will announce the content everytime current slide is changed even is user is elsewhere on the page
* when not in autoplay I also recommend to remove it and instead to move focus on the new active slide (or first new one in case of multiple items) when user activate .slick-prev / .slick-next or dots navigation button specially when there is focusable elements inside active slide

.slick-track : 
* remove role="listbox" because it's not relevant here. Sliders are not listbox see https://www.w3.org/TR/wai-aria-1.1/#listbox

.slick-slide : 
* remove role="option" because it's not relevant here as listbox is not relevant
* replace aria-describedby by aria-labelledby because label inside corresponding dot is more a label than a description
* remove this aria-describedby/labelledby when there is no dots navigation

.slick-dots : 
* remove role="tablist" because it's not relevant here. Dots navigation is not a tablist see https://www.w3.org/TR/wai-aria-1.1/#tablist
* on <li> inside it remove aria-hidden (because they need to be readable),  role="presentation" (because otherwise you loose <li> semantics, aria-selected (because not focusable), aria-controls (because not focusable)
* on button inside <li> remove role="button" (because useless), tabindex="0" (because useless)
* on button inside .slick-active add aria-current="true", this dot need also to visually be bigger than others otherwise there is information by color only
* by default I recommend to use "Slide x" where x= number as content inside button because it's more understandable for users than 1,2,3, etc
[...]

PR #3441, opened 2018-05-10
w3c accessibility refactor

kryp71c:
This PR implements the [accessible carousel suggested by w3c](https://www.w3.org/WAI/tutorials/carousels/working-example/).

The only differences to the w3c example are that the aria attributes for control remain in place, and the pause button is included in the controls list instead of the dot navigation list (this seemed to make more sense semantically). Based on [this article on heydonworks](http://www.heydonworks.com/article/aria-controls-is-poop), the aria-controls attribute is very poorly supported. Not sure whether it is detrimental to leave them in or not, so I left them in for now.

Working example can be found at http://jsfiddle.net/kryp71c/t2qffprm/2/
Only the last carousel in the example has the autoplay option turned on and therefore is the only one showcasing the pause button.
Apologies for the messy implementation of the fiddle, but I am not terribly familiar with this tool.

Note that when accessibility is turned off, the styles change (missing font makes the non accessible styles look off) as the accessible version gets a wrapper class "slick-accessible". When a11y is turned on, the focusOnSelect option only turns the focus on for the previous and next buttons (the focus always automatically changes when the dot navigation is used as per the w3c design pattern).

This PR is based on the discussion in https://github.com/kenwheeler/slick/issues/3268
One of the key differences to the discussion is the unchanged reading order. Seeing as the visual order is the same as the markup order, and w3c also suggests it this way, it seemed prudent to stick with it. For screen readers this might be more difficult to understand, but since the focus changes when the dots are used, it should still work.


kryp71c:
Hi @innocentiv,
Good catch. Interestingly, in the w3 example the playback control only changes state when the slide nav buttons are used, not when the arrow keys are pressed, or when the animation is paused because of focus or hover states. Probably best to move the state changes into the pause / play functions, and trigger pause when a dotnav item is pressed.

I also noticed the autoplay is not interrupted on hovering over the arrow keys. I will update the PR accordingly.

innocentiv:
Hi @kryp71c,
on a more thorough analysis, I found out some other issue:

- The pull request make use of two es6 feature: arrow function and default parameter. It's not really a big problem, for those who transpile their code, but I don't see the need to break ie11 support at the moment (no use of this inside the arrow function and undefined is already a falsy value)
- I think the dot content when multiple slide are showed is confusing. On the sixth example you provided, when I click on the arrows the screen reader read out "Slide 1 of 6" and "Slide 4 of 6". But when I click over the dots, the content is "1 of 6" and "2 of 6". I think that would be better to set the dot content to "1 of 2" and "2 of 2" to inform the user of the number of dots present.

Do you think you could address these issues as well?

kryp71c:
@innocentiv ah yeah, I'll look into the numbering and remove the es6 code, thanks.

kryp71c:
Finally had some time to implement the fixes.

@innocentiv perhaps take another look?

gabriel-cardoso:
Any updates on this PR ? Do you plan to merge it soon ?

kryp71c:
Hi @okeul,

Thanks, for pointing this out. I would agree with you on the focus, however the w3c design pattern leaves the focus on the prev / next buttons when navigating with them, and also uses aria live to announce the active slide (see [w3c working example](https://www.w3.org/WAI/tutorials/carousels/working-example/))

The rationale behind this PR was to bring the accessibility to a level that can be agreed upon, since there seem to be different opinions on what constitutes an accessible version. The w3c standard seemed like a good starting point for this.

Perhaps, the UX of the w3c recommendation is not ideal? It might be best to raise this issue with w3c first. They have a git repo for their accessibility tutorials: https://github.com/w3c/wai-tutorials

slick/slick-theme.css
```diff
@@ -1,204 +1,58 @@
-@charset 'UTF-8';
```
itanex: This is not a min file the structure o this file shoudl retain its new lines and human readable syntax

slick/slick.js
```diff
@@ -1706,9 +1788,16 @@
         _.focussed = false;
         _.interrupted = false;
 
+        if (_.options.accessibility && _.options.autoplay) {
+            _.$pauseButton.attr('data-action', 'pause')
+                .toggleClass('slick--playing slick--paused')
+                .find('.visually-hidden')
+                .html('Start Animation');
```
okeul: On this line, you should replace _start animation_ by _stop animation_.

For i18n, it's better to move this string in a variable : `.html(_.options.pauseLabel);`)

slick/slick.js
```diff
@@ -1694,6 +1769,13 @@
         _.autoPlayClear();
         _.paused = true;
 
+        if (_.options.accessibility && _.options.autoplay) {
+            _.$pauseButton.attr('data-action', 'play')
+                .toggleClass('slick--playing slick--paused')
+                .find('.visually-hidden')
+                .html('Start Animation');
```
okeul: For i18n, it's better to move this string in a variable : `.html(_.options.playLabel);`)

slick/slick-theme.css
kryp71c: @itanex thanks, will update.

slick/slick.js
kryp71c: @okeul thanks, and good catch. will update.

senhduong:
Hi, any update on merging this PR? Thx!

PR #3511, opened 2018-07-19
Implement beforeChange preventDefault support

boylett:
Continued from [this thread](https://github.com/kenwheeler/slick/pull/2104)

Works well and does not call the afterChange event

http://jsfiddle.net/132f87jL/


sonovice:
Any change of this getting pulled anytime soon?

Issue #2864, opened 2017-04-26
center cropped images


ig1na:
hello, 

the slick slider is working well, but when the window is resized, the slider properly resizes itself, the image is cropped (which is what I want) but only the right side is cropped, any idea how to crop the left and right sides?

thanks!

laurabennett:
Same issue.  I really want to center the image within its parent div but there doesn't seem to be an easy way to do that.

PR #3522, opened 2018-07-25
Make variableWidth working with vertical + center mode.

albancho:
Makes variableWidth working with vertical + center mode, as if it were a new "variableHeight" option.
Actually, the code adds to the list's top attribute the result of the "getLeft" function, but it miscomputed values from "left" et "width" slides attributes. Now uses "top" and "height" atributes if in verticalMode, so the list is correctly centered.

Actual slick with mis-calculation of top position: https://jsfiddle.net/alcho/vnL58e7y/
Pull-requested version with "variableWidth" option: https://jsfiddle.net/alcho/jbz230Lp/

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -1162,12 +1162,20 @@
 
             if (_.options.rtl === true) {
                 if (targetSlide[0]) {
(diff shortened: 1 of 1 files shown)


retober:
Works for me - and also fixes #2864 but the documentation should also be adjusted if this is merged.

PR #3614, opened 2018-10-22
Update jQuery dependency version

orditeck:


nick1190:
does slick support jquery 3.3.1?
why I can't work on Jquery 3.3.1
only support
 <script      type="text/javascript"     src="//code.jquery.com/jquery-1.11.0.min.js"    ></script>
    <script
      type="text/javascript"      src="//code.jquery.com/jquery-migrate-1.2.1.min.js"    ></script>

FreeFull:
@nick1190 
I tested it, and slick does seem to work with jQuery 3.3.1. Note that the example `index.html` file in this repo uses `$(document).on('ready', function() { .... })` which no longer is supported in jQuery 3. See https://jquery.com/upgrade-guide/3.0/#deprecated-document-ready-handlers-other-than-jquery-function

nick1190:
@FreeFull 


Thanks very  much


I change $(document).on('ready', function() { .... }) to $(document).ready(function(){.......})
and it 's work now

PR #3618, opened 2018-10-27
Fix grammar in readme, contributing and issue files

listopadov:


listopadov:
Fix flicker bug with Arrows in IE11

listopadov:
Fix selector into less file

PR #3627, opened 2018-11-06
Allow the 'appendDots' option to work with multiple elements properly

legroomdwyer:
Using 'appendDots' like this:
`appendDots: $('.dots')`
where there's multiple .dots elements or something like this:
`appendDots: $('.dots-1, .dots-2')`

The existing implementation already populated multiple dot containers and would change to the correct slide when they were clicked but it wasn't setting the .slick-active class on the appropriate li in each of the dot containers.

This change swaps jQuery's .eq() method for CSS :nth-child() so that it'll put .slick-active on the appropriate li in each dot container.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -496,7 +496,7 @@
 
             _.$dots = dot.appendTo(_.options.appendDots);
 
-            _.$dots.find('li').first().addClass('slick-active');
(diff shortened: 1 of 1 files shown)


s11e08:
Thx!

mister-cairns:
Thanks for this — very useful.

PR #3307, opened 2018-01-23
setSlidesClasses selector incorrect when index is 0

PaulBarrett79:
Mainly to fix #3271 where users are complaining about center mode jumping when going from last to first slide, I think there are other tickets where this may be the cause of jerky animations

This js fiddle shows the issue http://jsfiddle.net/paulbarrett79/5z11Lo7v/2/
This fix ensures the correct clone gets the slick-center class when going from last to first in center, infinite mode.

My fix is shown working here https://codepen.io/PaulBarrett79/pen/WdWYaO

PR #3638, opened 2018-11-22
Fixes slick-acitve class assignment with even number of slides and centerMode is active

ismailcherri:
> When enabling centerMode and choosing an even number of slides to show, the `slick-active` class is not properly assigned to the currently visible slides.

---

#Example jsfiddle:
###### `[  http://jsfiddle.net/ismailcherri/1ynwq34r/  ]`  

---

1. Create a slider with any number of slides  
2. Set the `centerMode` to `true`
3. Set the `slidesToShow` to an even number e.g. 4,6,8, etc

---

The `slick-active` class should be assigned to the visible slides only.

---

The `slick-active` class is assigned to the last slide on the right which is partially visible.  

---

#### More Details

* Which browsers/versions does it happen on? 
All browsers
* Which jQuery/Slick version are you using? 
Slick 1.9.0 and jQuery 2.2.0
* Did this work before? 
Not sure


PaulBarrett79:
I fixed this 9 months ago... #3307 

This desperately needs putting in the latest code, we now have multiple people fixing the same bug

ismailcherri:
Oh, I didn't see this, sorry.

PR #3674, opened 2019-01-02
Clone originalSlides in cleanUpRows to keep registered events

JPustkuchen:
When using unslick in combination with registered events (e.g. click on a slicked children element) even with jQuery .on() the events are removed. The reason is that empty() kills the events from the children before they are appended. By cloning them before, we keep the events. An example real world use case would be a modal registered on a slick element which doesn't work after unslicking anymore.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -822,7 +822,8 @@
         var _ = this, originalSlides;
 
         if(_.options.rows > 0) {
-            originalSlides = _.$slides.children().children();
+            // Get the original slides and clone them to prevent the events from being removed in empty() call below:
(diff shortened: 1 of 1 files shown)


JPustkuchen:
Any active maintainer? I guess there are some other issues based on this problem?

JPustkuchen:
The patch still works and is required. min.js should be generated, not corrected manually.

JPustkuchen:
Any chance to review this? ;)

JPustkuchen:
Would still be great to merge this to fix the described bug!

PR #3686, opened 2019-01-17
Fixes flicker at the end of rotation

### LINKED ISSUE #1890: Slide blinks at the end of infinite-rotation-transition

nchen63:
Fixes #1890 

It looks like the flicker was caused by 2 transitions applied at the same time. Spacing them out seems to fix the flicker.

Not sure what other side effects this may have, though.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -325,7 +325,7 @@
                         _.disableTransition();
 
                         callback.call();
-                    }, _.options.speed);
+                    }, _.options.speed * 1.5);
                 }
 
             }


andrewcbi:
have tested and this has removed the flicker, cannot find any errors caused by this change.

Issue #3355, opened 2018-02-28
Cursor hand


ChunkBytes:
Hi,

is there any reason why you use the value "hand" for the "cursor" css property? To find on the class "dragging" in the file "slick.scss/.less".

It's not a serious issue but it is not css valid and causes a error in the w3c css validator (http://jigsaw.w3.org/css-validator/validator). 

Thanks
Chunk

noobbbbb:
I am facing the same problem.
However, it may be necessary for the legacy IE.

rivajunior:
@noobbbbb Which versions of IE?

sarahmb83:
I am also facing the same problem.

ARXChrono:
> I am facing the same problem.
> However, it may be necessary for the legacy IE.

Nah it's not, its just a type mistake. Not sure if theres A existing PR for this. I'll make the change anyway but its uncertain how regular this repo is maintained so it may not be merged in.

caneta:
It would be nice to be fixed because it breaks W3C validation:
![image](https://user-images.githubusercontent.com/4066659/142600205-1b2ba254-092b-42e6-bd25-68eb08459247.png)

dkyuho:
Also having the same issue, will there be any updates on this?

<img width="474" alt="Screenshot 2022-11-29 at 3 59 04 pm" src="https://user-images.githubusercontent.com/103015432/204443273-8ca4573f-29ab-4886-8204-9f817b4c13f0.png">

Issue #2888, opened 2017-05-10
Cursor while dragging with mouse


szepeviktor:
Could we add `cursor: grabbing` while dragging?

szepeviktor:
I see https://github.com/kenwheeler/slick/blob/a737a4c98a4e342f33dcb548f54725c864b917e3/slick/slick.less#L28
but does not show up in Firefox on http://kenwheeler.github.io/slick/

szepeviktor:
I cannot find any `addClass('dragging')`

szepeviktor:
@leggomuhgreggo Could you take a look at it?

PR #3707, opened 2019-02-08
Fix: Cursor hand to grab

ARXChrono:
Resolves issue: https://github.com/kenwheeler/slick/issues/3355

 ### Description
Removed  **Cursor: hand;**
Added **Cursor: grab;** and **Cursor: grabbing;** 

However I'm unsure if dragging class is being toggled at all so it may be fine to remove that .slick-list.dragging style. Let me know if that is the case.

### Demo
http://jsfiddle.net/devshaun/fr8kgzwo/

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,3 @@
+*.scss
+*.less
+*.css
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -32,10 +32,12 @@
 {
     outline: none;
 }
-.slick-list.dragging
-{
+.slick-list.draggable {
     cursor: pointer;
     cursor: grab;
+}
+.slick-list.dragging {
+    cursor: grabbing;
 }
 
 .slick-slider .slick-track,
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -35,7 +35,7 @@
 .slick-list.dragging
 {
     cursor: pointer;
-    cursor: hand;
+    cursor: grab;
 }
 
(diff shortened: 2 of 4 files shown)


ARXChrono:
https://github.com/kenwheeler/slick/issues/2888

elliottmangham:
@ARXChrono this seems to still be an issue for me. I don't see any class added while dragging and therefore I cannot add the cursor style myself. It would be great if this was set by default with the option to remove.

ARXChrono:
> @ARXChrono this seems to still be an issue for me. I don't see any class added while dragging and therefore I cannot add the cursor style myself. It would be great if this was set by default with the option to remove.

Yeah actually, grab should be on draggable and dragging should have grabbing cursor.

andrewcbi:
@ARXChrono changing the css to this will make it work:


.slick-list.dragging, .slick-list.draggable:active:hover
{
    cursor: grabbing;
}

ARXChrono:
> 
> 
> @ARXChrono changing the css to this will make it work:
> 
> .slick-list.dragging, .slick-list.draggable:active:hover
> {
> cursor: grabbing;
> }

Oh sweet, you are right! Thought we had to rely on some JS from another PR.

endritvv (approved):

markvantilburg:
@ahmadalfy can you merge this can close these 3 issues below:

resolves: #3707
resolves: #2888
resolves: #3969

PR #3809, opened 2019-07-13
Fix #3560

### LINKED ISSUE #3560: Animation issues when going from last to first item.
When I try to animate the slide transition with some opacity, everything works fine except when I go forward from last to first element. Even when I go backward from first to last it works fine.

[.gif of the issue](https://imgur.com/a/SPuzZsx)

The gif is a bit more glitchy than the actual result but we can clearly see that when going from slide "4" to "1" the animation skips. But not when going from slide "1" to "4".

###### `[  https://jsfiddle.net/b5bqk68j/6741/  ]`  

1. Initialize a slick slider with the following config
    ```
    let sliderConfig = {
        centerMode: true,
        centerPadding: '400px',
        infinite: true,
        slidesToShow: 1,
        prevArrow: $prevArrow,
        nextArrow: $nextArrow
    };
    ```
2. Add transition to slide or element in the slide.
    in my case:
    ```
    .slick-animate {
        opacity: 0.5;
        transition: opacity 0.48s ease-in-out;
    }
    .slick-center {
        .slick-animate {
           opacity: 1;
       }
    }
    ```

3. Use any means of going forward through the slides and see the effect.

Every slides animates properly (like when going backwards).

When going from last to first the animation is glitching. (probably because of the slide cloning/switching)

explorador:
This is happening when number of items is not an even number.

When applying CSS transitions, slider is not running smoothly if it goes forward (from last to first slide). Works fine when going backwards (From first to last slide). Expected behavior: CSS transitions should run smoothly on both ways.

Without fix:
https://jsfiddle.net/explorador/y7r3p2nx/12/

With fix:
https://jsfiddle.net/explorador/dfs6gok3/11/

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2354,7 +2354,7 @@
                 if (index === 0) {
 
                     allSlides
-                        .eq(allSlides.length - 1 - _.options.slidesToShow)
+                        .eq( _.options.slidesToShow + _.slideCount + 1 )
                         .addClass('slick-center');
 
                 } else if (index === _.slideCount - 1) {


ahmadalfy:
This has been bogging me for a while, thanks. Will be ready for 1.9.1

PR #3817, opened 2019-07-24
Prevent refresh on an unslicked Slider

maxime-aknin:
If a manual $element.slick('unslick') is triggered on window resize, the timeout in `Slick.prototype.resize` is not cleared, thus causing arbitrary checkResponsive() to be triggered after the element was unslicked. Then `cleanUpRows` is called on an unslicked element which leads to slides being removed from DOM.


andrewcbi (commented):
this works correctly

PR #3818, opened 2019-07-24
Fix #3426

### LINKED ISSUE #3426: Custom paging stops working after version 1.8.0
Adding a custom paging stops working after version 1.8.0.

###### [Working CodePen](https://codepen.io/markushasselryd/pen/wmyKOd)
Working Codepen, but if I change to version 1.8.1 or higher, it stops working.

1. Set up slider as recommended in docs.
2. Add a data-title attribute on the slider items
3. Add customPaging like so:
` 
customPaging : function(slider, i) {
        var title = $(slider.$slides[i]).data('title');
        return '<a class="pager__item">'+title+'</a>';
    }
`

Dots should get the data-title information add replace the dots.

"undefined" gets output.

drdogbot7:
This fixes #3426 in my tests, but I don't know what other implications it might have. Data attributes were not being passed to the custom pager function.

This reverts a change from https://github.com/kenwheeler/slick/commit/34612b42641e8fd4250f70666a8da67eb624d002 which (apparently) fixed a bug with multiple rows.

Issue:
https://jsfiddle.net/drdogbot7/673w190p/

Fix:
http://jsfiddle.net/drdogbot7/h3b2jvgq/

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -559,7 +559,7 @@
         newSlides = document.createDocumentFragment();
         originalSlides = _.$slider.children();
 
-        if(_.options.rows > 0) {
+        if(_.options.rows > 1) {
 
             slidesPerSection = _.options.slidesPerRow * _.options.rows;
             numOfSlides = Math.ceil(


drdogbot7:
https://github.com/kenwheeler/slick/issues/3426#issuecomment-517830508

PR #3836, opened 2019-08-21
Correct version as defined in package.json

fridzema:


markvantilburg:
@ahmadalfy can you close this pull? We should update it to 3 or 3.7.1 since the official jquery project states `Only the latest version of jQuery is supported.`

PR #3876, opened 2019-11-07
add single classes to slick-dots children

Stoner-Bytelion:
Hi Ken,

Gotta say, I love slick. 
Thank your for all your work on this!

I noticed most elements have classes to style but the slick dots are the only ones left by the wayside.
This pull request adds single classes that follow the current naming convention in place to both `.slick-dot li` and `.slick-dot button`. My editor then trimmed trailing whitespace along lines 1022 and 1037.

Sincerely,
Bryan Stoner


markvantilburg:
Why would you need a class for every dot when the `ul` has a class set?

Issue #2401, opened 2016-07-03
Jumpy transitions when using waitForAnimate: false


andycansdale:
When waitForAnimate is set to false, the resulting transition is jumpy and disorienting when advancing various slides. It would appear that the ease is not being executed property when advancing to another slide before the previous transition has finished.

Recommend it be smoothed out if possible as in this example: http://www.onefinestay.com/homes/london/goodwins-court/

leggomuhgreggo:
Hey can you recreate in a reduced test case, like a jsfiddle or codepen? Thanks.

AllainPL:
It looks like an old glitch is still not fixed.
https://codepen.io/anon/pen/WXEPvQ

Try clicking fast on next/prev or random dots (buttons with numbers in example).

michaelCaleyWhaley:
This is still a problem

Issue #1737, opened 2015-09-16
slickNext (or previous) won't fire before transition is complete...


fusse:
Got a complaint from a client today that the slider doesn't work.

Turns out, the issue is that the slider won't advance to the next slide until the previous slide has fully changed. This becomes very obvious when using long slow transitions. In this case transition is two seconds... In that time customer manages to click on next arrow multiple times and complain that the slider doesn't work.

Try it: http://jsfiddle.net/4sktLkbg/2/ - try to advance to the next slide while the fade animation is still going. Doesn't happen...

Is this a bug or a "feature"? Cause I sure would like to make it possible to advance on click whenever. Even if the fade is cut and the next image is displayed in full.

kenwheeler:
set options.waitForAnimate to false

fusse:
Thank you so much for replying. The reason I missed this option is that I was reading the docs here:
http://kenwheeler.github.io/slick/ - waitForAnimate is not there. But it's here: https://github.com/kenwheeler/slick that I didn't think to read as thought these would match...

constantin-iliescu:
Hello,

I'm trying control the carousel with a jquery ui slider. waitForAnimate makes this reach the correct slide, but still, when you call the goTo method again, right before the animation completes, the slider kind of "jumps" directly to the new index instead of smoothly animating it.

Can you have a look here, please? 
http://codepen.io/anon/pen/qOXjRK

Thank you.

PR #3958, opened 2020-05-04
Fix: wait for animate bug

### LINKED ISSUE #607: waitForAnimate and fade bug
Hello,

me again.

I've found a new bug with your new feature waitForAnimate and slow fading.

If you fast click (multiple times) the dots before the animation is complete it sometimes jumps and not fades: http://jsfiddle.net/nveu9z8w/2/ (just click the dots a few times fast)

It seems that the fadings are queued and so happens that sometime all slides are about to fade out and  you only see the background-color. 

Could you please check this?

Thank you,
Pomm0

sasensi:
When `waitForAnimate` is set to `false`, if you trigger multiple consecutive navigations, you see a glitch at the end of the animation (the CSS transition is cleared by the first animation callback, while the last animation is still running).  
This makes sure that only the last callback is executed.  

Without fix:
http://jsfiddle.net/xidi2xidi/yzn9v5ej/6/

With fix:
http://jsfiddle.net/xidi2xidi/xkwgjd8e/1/

Closes #607 #1737 #2401

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -320,7 +320,11 @@
                 _.$slideTrack.css(animProps);
 
                 if (callback) {
-                    setTimeout(function() {
+                    if (_.currentTimeout) {
+                        clearTimeout(_.currentTimeout);
+                    }
+                    _.currentTimeout = setTimeout(function() {
+                        _.currentTimeout = null;
 
                         _.disableTransition();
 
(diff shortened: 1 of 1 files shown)


markvantilburg:
@ahmadalfy Slick might actually need this fix

PR #3979, opened 2020-06-13
Fix bug where single row slider is still generating extra row markup

iansvo:
Fixed options check to ensure extra row markup only gets created where there is more than 1 row.

Added extra class to the row element to make its purpose less ambiguous.

For several years now, I've noticed that a single row slider still generates this extra markup. Since the markup is just an empty div, its purpose is not exactly clear and thus has tripped up several developers I've spoken with.

Since the default row value is 1, this seems like a simple and appropriate change to make the option work as expected and have the resulting markup be more clear for the developer.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -71,7 +71,7 @@
                 pauseOnDotsHover: false,
                 respondTo: 'window',
                 responsive: null,
-                rows: 1,
+                rows: 0,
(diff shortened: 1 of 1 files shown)


hissssst:
I just don't understand how this component can be that much broken for 3 years

PR #4007, opened 2020-08-11
Feature/wwwisc 87

StephaniePoll:


StephaniePoll:
Note: Can't figure out how to minify the js file, and I'm a little hesitant to overwrite it with a different minifier. We use the non minified version in WWW, so maybe that isn't necessary?

markvantilburg:
This is also related to #4221

PR #4033, opened 2020-10-08
Accessibility: remove outline: none from themes

LuomaJuha:
This is a way to make this js more accessible as the outline is the most obvious way to tell if user has focused a certain button.


dieppon:
Please approve this commit, 

http://www.outlinenone.com

PR #4052, opened 2020-12-21
Fix index

khatrn:
Change image path and edit css in index.html


markvantilburg:
I can see the images should be replaced with other images, it would be better if it would be copies of the placeholder images instead.

markvantilburg:
Closed by pull https://github.com/kenwheeler/slick/pull/4335

PR #4091, opened 2021-05-03
Prevent invalid empty id attributes.

digitaldonkey:


marcoliverteschke:
For context: this pops up when you send your actual DOM after JS has been evaluated through a validator instead of just the initial source returned from the server. This might happen, for instance, when your site gets audited for accessibility.

Having empty IDs in the ”offstage” slides invalidates the markup, because if an ID attribute is present, its value must not be empty.

Ashwary-Jharbade:
Hi @kenwheeler ,
Could you please review this PR? Setting id to empty will cause cloned elements to lose CSS properties as well if applied via id. Please share your thoughts if you think it will have any negative impact.

markvantilburg:
This has another solution with setting a random value #4126

Issue #3799, opened 2019-06-21
Make Slick CSP compliant


ayumihamsaki:
I did a quick search and found this issue: https://github.com/kenwheeler/slick/issues/2399

To cut a long story short. All you need to do is remove the injected `style=""` and move the injected CSS into the external CSS file.

Would you be willing to update the code and adding some documentation for people.

ahmadalfy:
That’s interesting I never knew that this violates CSP. The problem is that slick needs to inject inline style because it calculates a lot of things on run-time. Is JSSOR CSP compliant? I would really love to investigate and fix that if it can be done.

ayumihamsaki:
@ahmadalfy yes jssor is fully compliant, have been using their code with a full csp policy for a few years now. All that was needed to get tehir code to pass was to remove the inline styles and replace them with class tags to an external css file.

ayumihamsaki:
> Is JSSOR CSP compliant?

Yes, have been using jssor for a few years with a fully strict csp policy. Three years ago I contacted their support and they gave me instructions how to make their slider fully csp compliant.

You can probably contact them and find out how they done it, because when I spoke to them they gave me more than enough background info on their code.

JeanSebTr:
I'll give a shot at a PR when time allows as our CSP reports are flooded by this, but I though I'd share a few details if someone want to fix this in the meantime.

CSP is not a monolithic thing, but different policies that can be used to restrict what can be done on the page. In that case, the absence of the `unsafe-inline` policy prevent setting the style attribute directly.

As was identified in #2399, it's caused by the reset of the saved style (that reset is also called when initializing slick which is probably when the CSP violations happen for most). I'm not sure what the best fix would be, but the simplest one would be to call `style.clear()` instead of setting the attr when `originalStyling` is empty. This should at least work for the cases where no inline styles were present on the carousel items.

JeanSebTr:
Looks like my brain invented style.clear, this doesn't exist (or more precisely is the clear css prop). This is the equivalent idea:

```
var style = this.style;
for(var i=style.length-1; i>=0; i--) {
    style.removeProperty(style.item(i));
}
```

Issue #2399, opened 2016-07-01
Content security policy


ShadowDancer:
> Slick slider is not CSP compliant https://en.wikipedia.org/wiki/Content_Security_Policy
====================================================================

No fiddle because it requires server headers

====================================================================

#### Steps to reproduce the problem

1. Enable CSP on your server
2. Add slick

====================================================================

#### What is the expected behaviour?

Beautiful and safe website

====================================================================

#### What is observed behaviour?

CSP stopped execution because of this:
```
        _.$slides.each(function(index, element) {
            $(element)
                .attr('data-slick-index', index)
                .data('originalStyling', $(element).attr('style') || '');
        });
```

```
        if(_.options.rows > 1) {
            originalSlides = _.$slides.children().children();
            originalSlides.removeAttr('style');
            _.$slider.empty().append(originalSlides);
        }
```

```
        if (_.$slides) {

            _.$slides
                .removeClass('slick-slide slick-active slick-center slick-visible slick-current')
                .removeAttr('aria-hidden')
                .removeAttr('data-slick-index')
                .each(function(){
                    $(this).attr('style', $(this).data('originalStyling'));
                });
```
- setting inline style!

====================================================================

#### More Details

- Which browsers/versions does it happen on?
All browsers supporting CSP
- Which jQuery/Slick version are you using?
Present in slick/slick.js in master branch
- Did this work before?
Nope?

kenwheeler:
Whats the fix here? Sanitizing the originalStyling data ?

ShadowDancer:
I'm suprised that noone in web knows about https://en.wikipedia.org/wiki/Content_Security_Policy
It's here for few years already, and it completly solves XSS problems.

Fix is to stop using inline styles .attr('style', something).

kenwheeler:
Yeah that's not gonna happen

kenwheeler:
Ok so it looks like using the CSS object model to store and reapply those styles is the actual fix. If anyone cares to PR that, I'll gladly take it.

ShadowDancer:
For what purpose you need inline styles?

kenwheeler:
For literally everything having to do with dynamically positioning elements?

kenwheeler:
After reading the spec, it's not about having inline styles it's about how they are applied. CSS object model is compliant"

ShadowDancer:
Well depends on definition of inline :) I was using terminology from spec.

[...]

PR #4100, opened 2021-05-14
Fix CSP issues by using CSSOM to reset styles when possible

JeanSebTr:
This fix Content-Security-Policy issues for the most common cases (see #2399 and #3799).

That most common case being using slick without having the style attribute set on individual carousel items so that the reset of slick can remove all CSS props using the CSSOM APIs.

I'm not sure if test cases can be added for this.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -889,7 +889,15 @@
                 .removeAttr('aria-hidden')
                 .removeAttr('data-slick-index')
                 .each(function(){
-                    $(this).attr('style', $(this).data('originalStyling'));
+                    var originalStyling = $(this).data('originalStyling');
+                    if (originalStyling) {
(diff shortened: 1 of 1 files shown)


kamodulin:
Hi @JeanSebTr! I wanted to do this on my end since I don't know how long it will be before this is merged. I was just wondering if I only have to modify slick.js or do I also have to change slick.min.js? Thank you!

JeanSebTr:
Hi @kamodulin !  It depends on how you're importing slick into your project. If you use `slick.min.js`, you'll have to build it and honestly, I did not manage to do so. I instead used [Accessible360/accessible-slick](https://github.com/Accessible360/accessible-slick) which has a more modern build process.

PR #4145, opened 2021-10-20
Fixed the single Dot and the extra div

somethumb:


markvantilburg:
@ahmadalfy Can you close this pull. This is already fixed with the latest build since the minified version was not updated.

PR #4146, opened 2021-11-02
Update scss url with interpolation

billyct:
## introduce

when I use vite, and do something with this file structure

```html
<!-- index.html -->
<script type="module" src="/nested/main.js"></script>
```

```js
// nested/main.js
import './style.scss'
```

```scss
// nested/style.scss
@import "slick-carousel/slick/slick-theme.scss"
```

## issue

vite will replace the url with `@return url(../node_modules/slick-carousel/slick/$slick-loader-path + $url);`

and cause `expected ")".` error

### DIFF
--- a/slick/slick-theme.scss
+++ b/slick/slick-theme.scss
@@ -26,7 +26,7 @@ $slick-opacity-not-active: 0.25 !default;
         @return image-url($url);
     }
     @else {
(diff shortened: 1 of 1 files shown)


markvantilburg:
@ahmadalfy This looks good to me

ahmadalfy:
Done

PR #4265, opened 2023-06-15
Fixes: SlickGoTo Event conflict with SlidesToShow and Doesn't change the slide to specified index

Arifursdev:


pablo67340:
Merged! Thanks for your contribution!

PR #4315, opened 2024-12-12
Update slick.css

HoosLab:
The CSS cursor property does not have a hand value. Instead, it has been updated to the grab value.

### DIFF
--- a/slick/slick.css
+++ b/slick/slick.css
@@ -35,7 +35,7 @@
 .slick-list.dragging
 {
     cursor: pointer;
-    cursor: hand;
+    cursor: grab;
 }
 
 .slick-slider .slick-track,


markvantilburg:
@ahmadalfy if changed in sass and less this could be merged.

Hand is not a valid value:
https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

If merged this should also be changed in:

https://github.com/kenwheeler/slick/blob/549ab0e87409fa6bfa40d53f149e6e3aec3e64c8/slick/slick.less#L30C1-L31C1
https://github.com/kenwheeler/slick/blob/549ab0e87409fa6bfa40d53f149e6e3aec3e64c8/slick/slick.scss#L30C1-L31C1

markvantilburg:
@ahmadalfy : Already fixed in this pull: #3707 , when that is merged this one can be deleted

ahmadalfy:
Done, I will close this one

PR #4331, opened 2025-06-23
Replace each occurrence of `.attr('id', '')` with `.removeAttr('id')`.

### LINKED ISSUE #4296: Empty id attributes are added to the slick item elements causing HTML validation error (valid-id)
When the Slick carousel is rendered the items have empty `id` elements. You can reproduce it with the jsfiddle you've provided in the issue template. It's an issue where HTML validation is needed to pass, see: https://html-validate.org/rules/valid-id.html#:~:text=Strictly%20the%20HTML5%20standard%20defines,%22%20becomes%20%23%5C%5C31%2023%20).

###### `http://jsfiddle.net/simeydotme/fmo50w7n/`  

Use Slick carousel as it's described in the documentation, nothing extra is needed.

Don't have empty `id` attributes, just simply remove the attribute if not needed.

Elements have empty `id` attributes.

#### Workaround

 I have the following workaround:

```
        $carousel.on("init", () => {
            $carousel.find("[id]").filter((_, element) => element.id === "").removeAttr("id");
        });
```

#### More Details

- Slick version: 1.8.1

sarahelsaig:
Fixes #4296

This PR replaces each occurrence of `.attr('id', '')` with `.removeAttr('id')`, as suggested by issue author:

> Don't have empty id attributes, just simply remove the attribute if not needed.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -2441,18 +2441,18 @@
                 for (i = _.slideCount; i > (_.slideCount -
                         infiniteCount); i -= 1) {
                     slideIndex = i - 1;
-                    $(_.$slides[slideIndex]).clone(true).attr('id', '')
+                    $(_.$slides[slideIndex]).clone(true).removeAttr('id')
                         .attr('data-slick-index', slideIndex - _.slideCount)
                         .prependTo(_.$slideTrack).addClass('slick-cloned');
                 }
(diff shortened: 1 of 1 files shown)


sascha-meissner:
+1 Sites are failing BITV accessibility tests because it errors duplicate id "" and invalid value for id ""

PR #4334, opened 2025-07-21
Update slick.js

markvantilburg:
Make sure all lines end with a ```;``` to comply with the code standard, and we can minify better.

image


ahmadalfy:
Sure

Issue #3697, opened 2019-01-31
Slick-dots not visible in safari and iphone 


mithunmaddur:
> short description of the bug / issue, provide more detail below.

====================================================================

###### `[  paste your jsfiddle link here  ]`  

use this jsfiddle to reproduce your bug: 
http://jsfiddle.net/simeydotme/fmo50w7n/ 
we will likely close your issue without it.

====================================================================

#### Steps to reproduce the problem

1. ...  
2. ...  

====================================================================

#### What is the expected behaviour?

...  

====================================================================

#### What is observed behaviour?

...  

====================================================================

#### More Details

- Which browsers/versions does it happen on?
- Which jQuery/Slick version are you using?
- Did this work before?

sir-red-dab:
I'm currently having the same kind of issue as you. Only I can't see the dots on any screen I use. When in inspection mode I can see they are there, but you can't change the styles to make them visible.

tisu19021997:
I'm having the same problem.

leko91:
In case someone stumbles upon this, I solved it by just giving `background-color: #fff` to the button.

josebrownie:
Hi,

There is any solution for this? I've the same issue only on iOS devices

Thanks

DC-Martin:
As @leko91 said, giving it a background color worked for me... if you have a different background color on the slides, then add that same color in rgba format and give it an opacity of 0.01 - that way the background on the dots wont be seen during the transition.   For example:
`.slick-dots li {
background-color: rgba(242, 243, 244, 0.01);
}`

PR #4343, opened 2025-10-16
Remove font smoothing (mostly unsupported css) 

markvantilburg:
Should also solve issues like: https://github.com/kenwheeler/slick/issues/3697

### DIFF
--- a/slick/slick-theme.css
+++ b/slick/slick-theme.css
@@ -74,9 +74,6 @@
 
     opacity: .75;
     color: white;
-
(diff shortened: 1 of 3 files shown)


markvantilburg:
@ahmadalfy can you merge this?

Looking at the documentation, it's not widely supported and since the "dot" is really small i doubt most visitors would even notice it.  
https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth

PR #4352, opened 2026-01-27
Using unitless 0 instead 0px values

shur:
In CSS, transform: translate3d(0px, 0px, 0px) and transform: translate3d(0, 0, 0) are functionally identical, as browsers interpret a unitless 0 as 0px in this context.

Can we use unitless 0 against 0px in translate3d(0, 0, 0) and translate() as a more modern and generally preferred approach because it is cleaner and shorter? Well, 0px and 0em are still 0.

### DIFF
--- a/slick/slick.js
+++ b/slick/slick.js
@@ -295,12 +295,10 @@
                     step: function(now) {
                         now = Math.ceil(now);
                         if (_.options.vertical === false) {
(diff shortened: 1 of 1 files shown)


markvantilburg (approved):
@ahmadalfy looks good to me

PR #4356, opened 2026-03-11
Clean up transform property in slick.scss

markvantilburg:
Removed vendor prefixes for transform property in slick.scss.


markvantilburg:
@ahmadalfy i forgot to include scss/less, can you merge this

PR #4357, opened 2026-03-11
Clean up transform property in slick.less

markvantilburg:
Removed vendor prefixes for transform property.


markvantilburg:
@ahmadalfy i forgot to include scss/less, can you merge this

Issue #3228, opened 2017-11-18
slick-theme.css @charset should use double quotes


FyiurAmron:
slick-theme.css @charset should use double quotes (https://github.com/kenwheeler/slick/blob/master/slick/slick-theme.css#L1)

====================================================================

#### What is the expected behaviour?

CSS should pass validation.

====================================================================

#### What is observed behaviour?

It doesn't, because, as per https://developer.mozilla.org/en-US/docs/Web/CSS/@charset , current @charset uses single instead of double quotes.

codepane:
Any update on this?

markvantilburg:
@ahmadalfy this is already fixed in the css file

PR #4366, opened 2026-09-14
Remove charset declaration from slick-theme.scss

### LINKED ISSUE #4013: /slick-theme.scss -- sass failed: Error: @charset may only be used at the root of a document.
When I use slick-theme.scss as a partial and @import it into my .scss structure, I get a build error:
`sass failed: Error: @charset may only be used at the root of a document. on line 1 of /code/php/stylelib/node_modules/slick-carousel/slick/slick-theme.scss`

If I remove the character set so does the error

1. add @import 'path/slick-theme';
2. compile

error occurs

#### More Details

I am using `npm install slick-carousel` --  Version: 1.8.1

markvantilburg:
Removed charset declaration from slick-theme.scss

fixes #4013

The parent file that imports it should declare the charset once at its root if needed. 
- Modern browsers default to UTF-8 encoding
- The Unicode characters work without it


markvantilburg:
also fixes #3228

